PR #8, opened 2019-04-01
Challenge 2

jbarrett:
- Writeup for challenge 1 copied from original gist
- Challenge 2 part 1 - Can `printf()` do it?
- Challenge 2 part 2 - I think this is correct :)


manwar:
Thanks for your contributions.

PR #12, opened 2019-04-03
Added solution for challenge-001.

adamcrussell:
Although I already made a "twitter submission" it seems the proper way to do it is via GitHub!


manwar:
Thanks @adamcrussell

PR #13, opened 2019-04-03
(re-)submission of weekly challenge solutions in suggested format (wk#002)

threadless-screw:


manwar:
Thanks @threadless-screw

PR #15, opened 2019-04-03
Challenge 002 & 001

andrezgz:
Hi Mohammad, I've resolved the challenges of this week and the previous one.
Is it OK or should I make a PR only with my Challenge 002 resolution?


andrezgz:
I've sent them in two separate PRs.

PR #19, opened 2019-04-04
challenge 1 + 2 perl 6 solutions and readme...

bagheera-sands:
Third attempt to create pull request - I've added perl 6 solutions (and fixed my perl 5 one...) the perl 6 solution uses currying to achieve the base conversion (even if it is a bit slow!)


manwar:
Thanks @bagheera-sands

PR #22, opened 2019-04-06
Solution challenge-002 Maxim Kolodyazhny

k-mx:
Hello! I already mailed request to join the challenge and here is my solution.


manwar:
Thanks @k-mx for sharing your solutions.
I liked the unit test. No one has ever submit before. 
Just a request, can you please save test script in the same folder rather than in the subfolder t/. 
The reason I am asking is, my parser may not like any subfolder in it. You could keep unit test in the same folder like ch-1.t or ch-2.t.
I will update the README with regard to the unit test.

k-mx:
Done

manwar:
Thanks @k-mx

PR #25, opened 2019-04-07
challenge-002 andrezgz fix ch-1

andrezgz:
Hi @manwar, reading the recap for ther first weekly challenge saw that "members can improve their solutions any time". Thanks for allowing it.


manwar:
Thanks @andrezgz

PR #31, opened 2019-04-08
Challenge 3 solutions

bagheera-sands:
No "golf"ing this time - but again looking at interesting Perl 6 methods - this time using lazy gather/take to create a "generator"....


manwar:
Thanks @bagheera-sands

PR #34, opened 2019-04-08
Some optimisations and some extra notes...

bagheera-sands:


manwar:
Thanks @bagheera-sands

PR #35, opened 2019-04-08
Rename challenge-003/khalid/ch-2.pl to challenge-003/khalid/perl5/ch-…

khalidelboray:
…2.pl


manwar:
Thanks @khalidelboray

PR #36, opened 2019-04-08
some naive perl5 solutions

lbalker:
I imagine the perl6 solutions will be radically different because of cleverness, if I get around to them!


manwar:
Thanks @lbalker

PR #46, opened 2019-04-12
imported my solutions plus a basic readme in the new perl5 directory

dcw803:
hi manwar, hope I'm doing this right, while I use git all the time, I've never forked or worked with branches and pull requests.  If I have done it right, this pull request should ?contain? my solutions to weekly challenge#3


manwar:
Excellent piece of work and well documented. Thanks for sharing @dcw803

PR #47, opened 2019-04-13
Have some iterator refinements

fjwhittle:
And blog metadata.  Is it against the spirit to update?


manwar:
Certainly not, in fact, we encourage such actions. Thanks for sharing your solutions.

PR #49, opened 2019-04-14
Fix typo in output, add optional timing output.

fjwhittle:
Somewhat embarrassed that I missed the – I guess that's what I get for not double-checking after selectively not commiting my timing code...
Figured I may as well include the timing code.


manwar:
Thanks @fjwhittle for your contributions.

PR #52, opened 2019-04-14
Add Perl6 solutions for Week 3

jmaslak:
Initial solutions in Perl 6.  Note that they don't have the normal comments - busy week!


manwar:
Thanks @jmaslak for sharing your solutions.

PR #57, opened 2019-04-15
changes to solution for challenge 2 - a bit longer but non-destructive!

bagheera-sands:
Just made some changes to look at performance issues - Perl 6 is slow Slow SLOW - not sure why tho!


manwar:
I love these small improvements as it makes understanding the code so easy about the improvements.
Thanks @bagheera-sands

PR #61, opened 2019-04-16
solutions for challenge 004

adamcrussell:


manwar:
Love reading your solution, keep up the great work.
Thanks @adamcrussell

PR #66, opened 2019-04-17
for ch-004 ch-1.pl

dmanto:
My solution to challenge-004 1
Based on the fact that a circle with radious N will have an area of π x N²
No square roots, no divisions until the very end, so is easier to estimate precision.
Relatively slow though :(


manwar:
@dmanto, it seems you misread the question. The challenge is the total number of PI digits should be same as size of your perl script. So if your perl script size is 10 bytes then the very same perl script should print 10 PI digits.

dmanto:
Oooops I thought script size was in lines :). So eight lines meant eight decimal places...
Sorry about that, I will think of some solution and come back...

manwar:
All the best :-)

dmanto:
please check this solution.
It's a 375 bytes perl 5 script that produces a 375 decimal places pi.
Hope this one is ok,
BR,

manwar:
Unique solution so far and it blew my mind initially.
Thanks @dmanto

PR #72, opened 2019-04-18
Submit solutions by Doug Schrag

dmaestro:
The high-precision sqrt / pi calculations are a module (Math::Precise) in progress :-)


manwar:
Thanks @dmaestro 
Somehow we are unable to send you weekly challenge, as the email address we have is rejecting every time.

PR #78, opened 2019-04-21
Comment about B::Deparse - musthave feature for perl developer

k-mx:
I thought it was worth mentioning B::Deparse.
P.S. Sorry for late PR.


manwar:
No need to be sorry.
You are encourage to make as many changes as you like before the deadline.
Thanks @k-mx

PR #81, opened 2019-04-21
my solutions to challenge004

dcw803:
Hi Manwar,

If I've done this right, please find my solutions to this week's problems - please tell me if I got it wrong, as I'm really not into all this "fork/pull request" stuff.

Working out how to refresh last week's fork was the hardest part, as I've not used forks before.  You might want to write out a simple guide for pull request/fork novices to say what to do THE SECOND TIME..  In my cloned fork from last week, I hoped "git pull" would freshen it up, but no chance.  I googled and found that I had to add your repo to my repo as the remote called "upstream", then "git fetch upstream", "git status" and "git merge upstream" (scary) and then "git push" back to my repo, before I could even start to add my solutions to challenge-004/duncan-c-white.  I didn't bother to create a branch this time, never really understood them:-)  All this github faff seems, frankly, more work than writing the damn Perl code, although I imagine it'll get easier.  If not, next time I might just email you a tarball, which takes a few seconds to make and send.

Anyway, that aside, really straightforward problems, although as generating digits of Pi is pretty dull, and the question didn't say "generate the same number of digiits of Pi as the size of your script", I decided the program should FETCH the first milion digits of Pi using LWP::Simple and store it locally, caching it in /tmp so that subsequent runs are quicker.

cheers
duncan

### DIFF
--- a/challenge-004/duncan-c-white/README
+++ b/challenge-004/duncan-c-white/README
@@ -3,13 +3,14 @@ as the size of your script. Say, if your script size is 10, it should
 print 3.141592653."
 
 Note that it DOESN'T SAY "calculate the same number of PI digits..." so
-I took the liberty to grab the over the internet rather than generate them
+I took the liberty to grab them over the internet (via a bitly link I set
+up to shorten the URL and hence the program) rather than generate them
 on the fly, because generating Pi is so dull.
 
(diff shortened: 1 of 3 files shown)


manwar:
Thanks @dcw803 
I will updated README with regard to working with previous fork.

PR #83, opened 2019-04-21
Another solution to challenge 004

gnustavo:
The new solution is more concise and easier to use, because it uses the
<> filehandle to read words from STDIN or from a sequence of files
provided as arguments.


manwar:
Thanks @gnustavo

PR #88, opened 2019-04-23
Perl5 solution - and selected optimizations...

bagheera-sands:
Written a few solutions this week - with iterations for improved performance....


manwar:
Thanks @bagheera-sands

PR #94, opened 2019-04-26
Added Perl6 Solutions for Challenge 5

Scimon:
Hi Mohammad, 

Here's my finished solutions for this weeks challenge in Perl6. I'm working on a blog post for it now which will also cover some of the false starts for part 2 and I'll send you a link to that when it's done.


manwar:
Thanks @Scimon

PR #96, opened 2019-04-27
Week 5 solutions in Perl 5 & Perl 6

jmaslak:


manwar:
Thanks @jmaslak 
Would you have time to blog about it?

PR #102, opened 2019-04-28
Solutions to 005 with built-in English dictionary.

bracteatus:
I closely followed Mark Senn's paper. The most significant difference is that I used Perl 5 instead of Perl 6. Other differences are
- The words are initialized within the script, instead of being read and prepared at runtime.


manwar:
Thanks @bracteatus

PR #106, opened 2019-04-29
Jmaslak week 6, problem 2

jmaslak:
Perl 5 & 6 solutions attached.

You linked to https://en.wikipedia.org/wiki/Landau%E2%80%93Ramanujan_constant which is the Landau-Ramanujan Constant.

That's not the same thing as the Ramanujan's Constant, which you can find here:
https://en.wikipedia.org/wiki/Heegner_number#Almost_integers_and_Ramanujan's_constant

Since the Ramanujan's constant is fairly easy to calculate to 33 digits of precision, I solved for that.  The Landau-Ramanujan Constant is not easy to calculate to 33 digits of precision!  Please let me know if I solved for the wrong constant.

### DIFF
--- a/challenge-006/joelle-maslak/perl5/ch-2.pl
+++ b/challenge-006/joelle-maslak/perl5/ch-2.pl
@@ -6,7 +6,7 @@ use warnings;
 use bignum;
 
 #
-# Copyright © 2019 Joelle Maslak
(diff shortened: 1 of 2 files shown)


manwar:
I have updated the link, thanks @jmaslak

PR #118, opened 2019-05-04
Challenge #006 - ch-1.p6 solution

threadless-screw:


manwar:
Thanks @threadless-screw

PR #130, opened 2019-05-09
challenge-007 andrezgz solution

andrezgz:
Hi @manwar, this week I've only managed to solve the first challenge. The other one I felt it was out of my league.


manwar:
Thanks @andrezgz

PR #138, opened 2019-05-12
Challenge 007

adamcrussell:
Solution for challenge 007.


manwar:
Thanks @adamcrussell for making it on time :-)

PR #139, opened 2019-05-12
Challenge 7 problem 1 by Jaldhar H. Vyas

jaldhar:
Owing to a severe lack of time, only the first problem today.  But I will try to do the second one tomorrow.


manwar:
Thanks @jaldhar

PR #140, opened 2019-05-13
Challenge 8 - plus updated notes on other challenges (with timings)

bagheera-sands:
Not sure if using the Mersenne route can be considered cheating here! But works fast even for large values of "N" - on a not very powerful VM - manged to get perl6 timings up to the 25th perfect number..

Getting first 25 took approx 30 minutes (first 20 about 20 seconds). The largest of which has over 13,000 digits....


manwar:
Thanks @bagheera-sands

PR #142, opened 2019-05-13
Perfect numbers 

Scimon:
Good morning. I'm hoping I'll be a bit more chirpy this week and will blog this one.

Plus I think the second test should be done soonish.

Not sure about the API thing.


manwar:
Thanks @Scimon

PR #147, opened 2019-05-14
Joelle's Solution for 8.3 in P6

jmaslak:


manwar:
You are star, my friend. Thanks for sharing your solution :-)

PR #150, opened 2019-05-15
challenge-008 andrezgz solution

andrezgz:
Hi @manwar, this week I was able to solve them both again :)


manwar:
Thanks @andrezgz

PR #154, opened 2019-05-16
Update readme to include timings for up to the 25th perfect number

bagheera-sands:


manwar:
Thanks @bagheera-sands

PR #158, opened 2019-05-18
my solutions to ch-1 & 2 in perl5

dmanto:
sorry didn't have the time for ch-3

I would like to take advance of this space to propose a problem to be included in a future challenge:

> Write a function, 'miner', whose first argument is a number greater than one, and then a list of strings, which will be lines of text. The function will return the same list of strings, followed by a line with only a number, so that the SHA256 hash of the whole returned list joined by "\n" end of lines will give a hash starting with "n" (or more) binary 0 positions, where "n" is the first argument mentioned.

The only perl module that should be used is Digest::SHA module (in perl5, in perl6 you may suggest an equivalent module I guess)

For example:

> miner(14, 'This', 'are my', 'test lines');

should return a list like:

> 'This', 'are my', 'test lines', '14131'

You can check it (having openssl installed, for example) with:

> $> echo -n "This\nare my\ntest lines\n14131"|openssl dgst -sha256

That will return:

> 0003d4db115adfabb5b77588ff9ea324f5a6d57481cec7b39d5fe5cfd9867d73

and, as you can see, starts with 14 binary zeros 0000 0000 0000 0011 ....

Hope you like it!
BR,

### DIFF
--- a/challenge-008/daniel-mantovani/perl5/ch-1.pl
+++ b/challenge-008/daniel-mantovani/perl5/ch-1.pl
@@ -1,0 +1,52 @@
+# Write a script that computes the first five perfect numbers.
+# A perfect number is an integer that is the $sum of its positive proper divisors (all divisors except itself).
+# Please check Wiki for more information. This challenge was proposed by Laurent Rosenfeld.
+
+use strict;
+use warnings;
+use v5.10;
+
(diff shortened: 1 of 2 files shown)


manwar:
Thanks @dmanto

PR #162, opened 2019-05-20
Added challenge 1

Scimon:


manwar:
Thanks @Scimon for the magical solution.

PR #187, opened 2019-05-28
I'd done this before for Exercism 

Scimon:
but I've tweaked my result with a unicode option

(Not using all the unicode points though, gets a bit weird for some of them).


manwar:
Thanks @Scimon

PR #188, opened 2019-05-28
Now reads Roman numeral strings

Scimon:
Can handle both ASCII and unicode numeral strings. Including the special unicode character for sopme of the numbers between 1 and 12.


manwar:
Thanks @Scimon

PR #189, opened 2019-05-28
Support larger numbers and include script with arabic numerals

mienaikage:
Will eventually get round to the decoding


manwar:
Thanks @mienaikage

PR #199, opened 2019-06-01
Solution for week 10 challenge 1 in P5 & P6

jmaslak:


manwar:
Thanks @jmaslak 
I was getting nervous not hearing from you yet ;-)

PR #10, opened 2019-04-02
Solution submission weekly challenge (#2)

threadless-screw:

PR #201, opened 2019-06-01
Solve #010

choroba:
The Wikipedia page of Jaro-Winkler is a bit unclear (reflected also in
its Talk page), so I settled for a CPAN solution.


manwar:
Thanks @choroba 
Would there be blog on the way? ;-)

choroba:
Probably, but later (Sunday or next week). I don't have enough time at the moment :-)

PR #205, opened 2019-06-02
solutions for challenge 010

adamcrussell:
I am still writing up my blog but it should be done by the deadline! link in blog.txt is valid and be the location of the completed blog.


manwar:
Thanks @adamcrussell

PR #214, opened 2019-06-03
Week 11 challenges

Scimon:
Here we go. I think I got them both OK.

Rational numbers for the win in the case of challenge 1.


manwar:
Thanks @Scimon 
I fall in love with Perl6 every time I read your code.

PR #231, opened 2019-06-09
solution for challenge 011

adamcrussell:
thank you as always for administering the weekly challenge!


manwar:
Thanks @adamcrussell

PR #233, opened 2019-06-09
Update ch-3.pl

PerlMonk-Athanasius:
Redacted API KEY !


manwar:
Thanks @PerlMonk-Athanasius

PR #234, opened 2019-06-09
hi Manwar, please find my solutions to this week's challenges, and belatedly by solutions to a couple of previous ones..

dcw803:
hi Manwar,

Please find my solutions to this week's challenges, and belatedly by solutions to a couple of previous
ones for info.  In challenge 8 I did them on time, thought I'd committed them, only later realised that
I'd completely forgotten the whole "pull request" part:-)

In challenge 10 I did the roman numerals part (I liked that) on time, but on the Jaro-Winkler distance
I run out of time because I found the Wikipedia description of the Jaro distance utterly confusing,
later the Sunday evening I resorted to looking at other implementations on Rosetta stone in order to
understand how to calculate the Jaro distance was.  Having got it working about 8pm, I tried to submit
them but got hopelessly confused while git merging the upstream changes in with mine and giving up
in disgust.  When it takes longer to try to submit something than it does to solve the problem, I give
up:-)  Can't I just email a tarball, takes 30 seconds?

Anyway, this week, with more time in hand, I think I fixed whatever screwup I got into last week, so
have included challenge 10's solutions anyway, and I've definitely done this week's - rather trivial -
challenges.  The first one - fahrenheit(c)==c is actually a matter of calculation (solve 9/5c+32==c, the
answer is -40, I was very tempted to submit "print -40" as my solution) but I implemented it as a "find c
such that abs(fahr(c)-c))<epsilon" solver.

The second one - the identity matrix - was utterly trivial, I wasn't sure whether to compute it internally,
as a 2-D array (as we would do if it was one operation among many in a Matrix class), and then print
it out (using that class's stringification method), but as the question didn't say (or specify the output
format) I just printed out a 2D block of 0s and 1s.


manwar:
Thanks @dcw803

PR #260, opened 2019-06-16
Create ch-2.p6

threadless-screw:
It works, but something (in particular the length of the code) tells me this is not the Perl way of getting this job done... I'm gonna have a look at the blogs for inspiration :-)


manwar:
Thanks @threadless-screw

PR #262, opened 2019-06-16
fjwhittle ch-3.p6 fixes after testing.

fjwhittle:
My API key came through so I did some testing, and found I had to fix up my script.


manwar:
Thanks @fjwhittle

PR #263, opened 2019-06-16
Work with actual API, not documented API!

jmaslak:
The documentation for the API's JSON output is just plain incorrect.  Now that I have actual access to the API, not just sample output from the API docs, this will work with the real API.


manwar:
Thanks @jmaslak

PR #270, opened 2019-06-17
further optimisations

joetym:


manwar:
Thanks @joetym 
I like the way you work on optimization.
Why don't you tell us how you progress through the optimization. It would be good read for others.

joetym:
Thanks, sure I can do that. I'm a bit confused why nobody else is using that module (Math::Prime::FastSieve). Is it not in the spirit of the challenge to use modules that already have the problem solved ?

PR #274, opened 2019-06-17
Jtym challenge12

joetym:


manwar:
Thanks @joetym 
I like the way you improve your code bit-by-bit. :-)

PR #282, opened 2019-06-20
Challenge 13 by Jaldhar H. Vyas: BUG FIXES

jaldhar:
That's what I get for doing last minute edits without tests :-)


manwar:
Thanks @jaldhar

PR #293, opened 2019-06-23
Challenge 13 by Jaldhar H. Vyas: blog post

jaldhar:
I've been meaning to blog for some time and now I've finally gotten around to it.  This is the post for the current challenge.  I've also written posts for all the previous challenges but they are not ready yet.  (Of course I have to write my own blog script first!)  The appearance will change a bit but the URL should remain stable.


manwar:
Thanks @jaldhar and congratulation for the first blog to Perl Weekly Challenge.

PR #295, opened 2019-06-24
Challenge 014

rvandam:
I'm new.  I also sent an email to be added but perhaps this is simpler.


manwar:
Welcome on board Robert.
Thanks for sharing your solutions :-)

PR #296, opened 2019-06-24
Challenge 1. Van Eck's Sequnce

Scimon:
Here you go. I'll take a look at the next one in a bit when my head stops aching.


manwar:
Thanks @Scimon

PR #303, opened 2019-06-25
solutions to challenge 14

waltman:


manwar:
Thanks @waltman

waltman:
You're very welcome. I guess this means it worked? :)

PR #307, opened 2019-06-26
Add link to Choroba's blogpost and the solution to the extra task

choroba:
Also, remove unneccesary condition.


manwar:
Thanks @choroba

PR #330, opened 2019-07-02
Finished this weeks, finally submitted some old ones I did

rabbiveesh:
I think I may actually write a blog post about this week, now that I'm back into things.
This is really fun, thank you so very much!


manwar:
Welcome back and thanks for sharing code.

PR #350, opened 2019-07-07
solutions for challenge 015

adamcrussell:


manwar:
I was waiting for your solution patiently ;-)
Your blog is going to be missed out in next week newsletter, as I already sent the list to the chief editor  :-(
Thanks @adamcrussell

PR #355, opened 2019-07-09
Roger's solution. I hate everything to do with Bitcoin.

Firedrake:
See Attack of the 50 Foot Blockchain for why.


manwar:
I loved both your solutions. You are simply genius. I will check your code one more time to understand the details.
Thanks @Firedrake

PR #360, opened 2019-07-11
challenge-016 andrezgz solution ch-1

andrezgz:
Hi @manwar, I've only managed to solve the first challenge of this week


manwar:
Thanks @andrezgz

PR #363, opened 2019-07-12
Solutions to challenge 016 problem 1 and 2 in perl6 by Noud

noudald:
Problem 2 was a difficult problem! It took me some time to find a good solution for problem 2.


manwar:
I am happy you got it at the end. Excellent.
Thanks @noudald

PR #364, opened 2019-07-13
ruben/drclaw solutions to W16 CH1 and CH2. P5 and P6

drclaw1394:
First time using bit manipulation in perl/perl6... I got there in the end!


manwar:
Thanks @drclaw1394

PR #369, opened 2019-07-14
Pyrrhlin pwc

myrrhlin:
yeah, i'm late for week 15.  but it's cool.


manwar:
It seems you didn't see my reply. I had already added it.
Never mind, good to see you using PR for submitting solutions.

Thanks @myrrhlin

PR #370, opened 2019-07-14
Add solutions of 016 by E. Choroba

choroba:


manwar:
Thanks @choroba 
Most nervous week, as regulars kept it last minute. :-)

PR #376, opened 2019-07-14
imported my solutions (nb: I really didn't solve ch-2.pl myself, see README)

dcw803:
The first question was beautifully clear, and a fun question.

But the Bitcoin address decoding question was horrid - I got rather bogged down in understanding how Bitcoin addreses were encoded, after reading various web pages starting with the wiki page you suggested, but it really wasn't clear what the algorithm was, until I found a reference implementation already written in Perl 5.

I simply couldn't be bothered to rewrite this code, after all I wouldn't have known what code to write to solve the problem without reading the reference implementation, which meant that I'd already seen a perfectly good solution.  So I nicked it, and adapted it slightly.


manwar:
Thanks @dcw803

PR #377, opened 2019-07-14
Challenge 16 Problem 1 by Jaldhar H. Vyas

jaldhar:
I've been travelling all weekend so I've had even less time than usual.  I've managed to complete the first task however.


manwar:
Thanks @jaldhar 
I am glad you made it :-)

PR #379, opened 2019-07-15
Ackermann thing

Scimon:
Sorry I seem to have messed up the git history a bit.


manwar:
Thanks @Scimon 
I have added your solution manually.

PR #380, opened 2019-07-15
Regexes were perhaps not the ideal tool, but.

Firedrake:


manwar:
I was kind of expecting you to be the first as far as Perl5 solutions are concerned.
And you didn't disappoint me.
Thanks @Firedrake

PR #381, opened 2019-07-16
all but the blogging and MAYBE the API challenge. Plus profile pic.

jacoby:


manwar:
I loved your clean regex, so neatly arranged. Thanks @jacoby for sharing solutions and photo.

PR #386, opened 2019-07-18
challenge-010 andrezgz minor fix for ch-1

andrezgz:
Hi @manwar, I've fixed the argument handling on a sub call.


manwar:
Thanks @andrezgz

PR #410, opened 2019-07-21
imported my solutions for this week

dcw803:
Hi Manwar,

Think I got these working, although neither question really inspired me.


manwar:
We will try to do better next time, sorry to disappoint you.
Thanks @dcw803

PR #412, opened 2019-07-22
Solutions to #18

Firedrake:
In both cases there are CPAN implementations that will do _most_ of the work… but that wouldn't have been as much fun.


manwar:
Thanks @Firedrake

PR #413, opened 2019-07-22
Longest substrings challenge

Scimon:


Scimon:
I've added the challenge two as well. It's got Perl6 kebab-case names AND Perl5 snake_case names. I might look at making it an Iterable. So you can just do `for $queue {}`

manwar:
Thanks @Scimon

PR #418, opened 2019-07-25
Add solutions to 018 by E. Choroba

choroba:
There are two different files for the first task, one of them searches
the longest common substring naively, the second one uses a suffix
tree.

The priority queue is also implemented in two different ways, but in
one file only: using an array or a heap.


manwar:
Thanks @choroba

PR #420, opened 2019-07-26
Fix solutions to 018 by E. Choroba, add a blog post (part 1)

choroba:
- small errors in the code translated from Java
- use a counter instead of a float priority


manwar:
Thanks @choroba

PR #435, opened 2019-07-28
Perl 6 solutions for task 1 and 2 of challenge 018 by Noud

noudald:
Thank you for organizing another interesting challenge. Here are my solutions for task 1 and 2. I'm looking forward to the solutions of the other participants.


manwar:
Thanks @noudald

PR #457, opened 2019-08-02
Challenge 019, Task #1 and #2

bracteatus:
This week's solutions make use of the PERL standard library and standard UNIX programs to speed up the development time.


manwar:
Thanks @bracteatus

PR #460, opened 2019-08-03
Adding Perl6, because I can.

Firedrake:


manwar:
You can do Perl5 as well ;-)
Thanks @Firedrake

PR #468, opened 2019-08-04
Solutions wk19ch1-2 in Perl6

threadless-screw:


manwar:
Thanks @threadless-screw

PR #481, opened 2019-08-06
challenge-020 andrezgz solution

andrezgz:
Hi @manwar, my solutions for the first 2 tasks (plus a one-liner for the first one)


manwar:
Thanks @andrezgz

PR #494, opened 2019-08-10
Solutions to challenge 20 problem 1 and 2 in Perl 6 by Noud

noudald:
I'm proud of my solution for problem 2. For the problem I create a lazy list with all Amicable Numbers. Moreover I use an anonymous cache handle to speed up the calculations. Making it possible to compute the first three Amicable Numbers within a couple of seconds. (on my laptop at least)

I'm looking forward to the challenges of next week!


manwar:
Thanks @noudald

PR #501, opened 2019-08-11
Checked in my (dcw)'s solutions for this week

dcw803:
Hi Manwar,

Two very simple but nicely specified problems this week.  A pleasure to do them both.

cheers
duncan


manwar:
I simply love your comments. It means a lot coming from you. I must be doing things right recently.
Thanks for sharing your solutions, @dcw803

PR #506, opened 2019-08-12
andrezgz challenge-020 fix challenge url in comment

andrezgz:
I've copy-pasted the scripts from the previous challenge as a baseline and forgot to adjust the url


manwar:
Thanks @andrezgz

PR #507, opened 2019-08-12
Added solutions to Challenge 21

davorg:


manwar:
Pleasantly surprised to see your solution. Your is the first solution received.
Thanks @davorg

PR #519, opened 2019-08-17
Solutions to challenge 021 problem 1 and 2 in Perl 6 by Noud

noudald:
* Problem 1 has been solved with continuous fractions. One of the fastest ways for approximating Euler's number.
* It's probably possible to make Problem 2 more efficient. I'm looking forward to see how others solved this problem.


manwar:
Thanks @noudald

PR #528, opened 2019-08-18
Challenge 21 by Jaldhar H. Vyas  (except blog and p6/ch-2)

jaldhar:
Ran out of time again :(  The Perl6 version of challenge 2 and the blog entry will be along shortly.


manwar:
Thanks @jaldhar

PR #531, opened 2019-08-20
Added template and perl5 solution for challenge #1

Doomtrain14:


manwar:
Thanks for your first contribution.

PR #542, opened 2019-08-22
Add solution to 022 by E. Choroba

choroba:
Sexy Prime Pairs uses the prime module I've already used in two
previous challenges, 012 and 015.

LZW works according to the example in the Wikipedia page, i.e. only
works for capital letters. It can be easily extended to handle any
other characters, though.


manwar:
Thanks @choroba

PR #557, opened 2019-08-25
finally got the LZW stuff working (with 5 mins to spare:-)) after…

dcw803:
…having difficulties getting text->encode->decode->sametext for some texts..

Not sure the wiki page's description of the DECODING was really clear enough.. but got it working in the end..


manwar:
You made it :-)
Thanks @dcw803

PR #560, opened 2019-08-26
challenge-023 andrezgz solution

andrezgz:
Hi @manwar , this week my solution has arrived early! Hope I've understood correctly the first task :)


manwar:
Well I don't expect you to misread the task ;-)
Thanks for sharing your solutions @andrezgz

PR #565, opened 2019-08-27
Move the test for n into the parameters

Scimon:
Sorry @manwar I thought of a neater way of testing my value.


manwar:
:-)
Thanks @Scimon

PR #569, opened 2019-08-29
Solutions for challenge 23 tasks

randyl:
This week's tasks were easier, which was very welcome after the tough (but fun) compression algorithm last week! Thanks, @manwar.


manwar:
Glad you found task easy. I can't thank enough you for your participation week after week.
Thanks @randyl

PR #577, opened 2019-08-31
Solutions to challenge 023 problem 1 and 2 in Perl 6 by Noud

noudald:
I had not time last week to do challenge 022. But this week I could find some time again to work on the Perl Weekly Challenges. Here are to solutions for challenge 023 in Perl 6.


manwar:
Welcome back @noudald 
It is always pleasure reading your code.

PR #589, opened 2019-09-02
Apparently an empty file runs fine in both perl5 and perl6

Scimon:


manwar:
I am enjoying this task :-)
Thanks @Scimon

PR #599, opened 2019-09-05
challenge-024 andrezgz solution

andrezgz:
Hi @manwar these are the solutions for the first 2 tasks of this week challenge. I've added some txt files with Aesop's Fables to use them with ch-2.pl


manwar:
Thanks @andrezgz

PR #602, opened 2019-09-07
Solutions for challenge 023 problem 1 and 2 in Perl 6 by Noud

noudald:
I understand the idea behind problem 1. Unfortunately Rakudo doesn't complain often. It's not really a clang for perl6. ;)

Looking forward to the new challenges next week!


manwar:
Thanks @noudald

PR #608, opened 2019-09-08
Add a link to E. Choroba's blog post

choroba:
plus the shortest script that doesn't do anything special but doesn't
throw an error either.


manwar:
Thanks @choroba

PR #614, opened 2019-09-10
Added perl5 solution for task#2

Doomtrain14:
Added perl5 solution for task#2
Updated solution for task#1


manwar:
Thanks @Doomtrain14

PR #627, opened 2019-09-15
Perl6 solution to challenge 1 - dogslow, but it works

threadless-screw:


manwar:
Thanks @threadless-screw

PR #636, opened 2019-09-16
Busy then ill last week but here's a start for this week.

Scimon:


manwar:
Sorry to hear that. Hope you are feeling better now.
Thanks @Scimon

PR #637, opened 2019-09-16
Think this does the job.

Scimon:
Note the degrees version rounds to 0.01deg as deg -> rad -> deg conversion tends to add some float errors.


manwar:
Thanks @Scimon

PR #638, opened 2019-09-16
Add solution to 026/1 (stones & jewels) by E. Choroba

choroba:
Introduction of a new secret operator, +=()=, that could be called
"Voyager approaching Saturn".


manwar:
Cool tricks :-)
Thanks @choroba

PR #657, opened 2019-09-22
Challenge 25 by Jaldhar H. Vyas: all tasks including an update for Pe…

jaldhar:
Here's the rest of challenge 25.


manwar:
Thanks @jaldhar

PR #667, opened 2019-09-24
challenge-027 andrezgz solution

andrezgz:
Hi @manwar, hope the second task meets the requirement


manwar:
Interesting hack ;-)
Thanks @andrezgz

PR #670, opened 2019-09-26
Doesn't *quite* hit the spec but does give you objects you can easily…

Scimon:
… use and keep track of the history off


manwar:
Thanks @Scimon

PR #680, opened 2019-09-29
Solution for 27.1 in Perl 5, minor code fixes for Perl 6 version

jmaslak:


jmaslak:
Don't merge, there's some screwups here.

PR #681, opened 2019-09-29
Solution for 27.1 in Perl 5, minor typo fix for Perl 6

jmaslak:


jmaslak:
Don't merge, there are some screwups here.

PR #685, opened 2019-09-29
imported my challenge-027 solutions..

dcw803:
Hi Manwar,  Had a go at both challenges, used Gnuplot (optionally) to visualize the results in both occasions.  The second challenge was rather ill-defined, but I've interpreted it in a particular way, a sequence of "x += 10" type input lines, then I track all values of each variable over simulated time (assuming line N happens at time N), then display the results.


manwar:
Thanks @dcw803

PR #702, opened 2019-10-01
Solved challenge 1

rabbiveesh:
I actually just learned about the character class `[[^:ascii:]]`, so it was fun to use it!


manwar:
Thanks @rabbiveesh 

Have you joined Hacktoberfest yet?

rabbiveesh:
Of course.
Why do you think I made a pr with only one solution 😉

PR #28, opened 2019-04-07
Perl Weekly Challenge 002

AlexDaniel:

PR #709, opened 2019-10-02
Feng Chang's solution for challenge #28

seaker:


manwar:
Welcome back, we missed you ;-)
Thanks @seaker

PR #725, opened 2019-10-06
incorporated my solutions.. ch-1.pl is ridiculous, but

dcw803:
ch-2.pl is a rather cool 7-segment led based digital clock (that resizes properly too)


manwar:
Thanks @dcw803

PR #732, opened 2019-10-08
Solutions to challenge 29 tasks 1 and 2 in Perl 6 by Noud

noudald:
I like the tasks of this week. I found concise solutions to both of the tasks. I'm especially impressed with how much ease it is to call c functions within perl6. Thank you!


manwar:
Thanks @noudald

PR #737, opened 2019-10-09
challenge-029 andrezgz solution

andrezgz:
Hi @manwar, the C task was completely new to me but I've learned tons of interesting stuff. The other one was more like try/error. Sure, I didn't catch all the "extreme cases" but I'm confident I've got a decent solution


manwar:
Thanks @andrezgz

PR #757, opened 2019-10-13
imported my solutions..

dcw803:
two interesting questions, made me think.  Haven't used XS or Inline::C before, gave the latter a go  For the first question, did it from whole cloth. Tricky!


manwar:
Thanks @dcw803

PR #758, opened 2019-10-13
added Mark Senn's solution to PWC 029 task 1 in Perl 6

mark-senn:
If I've done the git stuff right this is ok to merge.


manwar:
Thanks @mark-senn

PR #762, opened 2019-10-14
minor corrections

threadless-screw:
When staring at the ceiling last night, two little things kept me awake :-)


manwar:
Thanks @threadless-screw

PR #767, opened 2019-10-14
Challenge 30

bagheera-sands:
Haven't written a solution recently - but thought I'd have a "swing" at these - sorry the solutions are short - thought I'd see how easy these were to Golf....

Total length of shell scripts is just 116 bytes! and that includes 2 x 9 bytes for the perl -E '' parts.


manwar:
Welcome back and thanks for sharing @bagheera-sands

PR #769, opened 2019-10-14
Fixed order of challenges

bagheera-sands:
Foo messed up order of challenges....


manwar:
Thanks @bagheera-sands

PR #812, opened 2019-10-20
Solved challenge 30

Leoltron:


manwar:
Thanks @Leoltron and welcome on board.

PR #816, opened 2019-10-21
Challenge #031

davorg:


manwar:
Thanks @davorg.
You are the first to submit the solutions.

PR #817, opened 2019-10-21
Added perl5 solution ch#31-1

Doomtrain14:


manwar:
Thanks @Doomtrain14 for being so quick to respond.

PR #819, opened 2019-10-21
Challenge 1. Test a bunch of random divisions to see if they are safe…

Scimon:
… without looking at the divisor


manwar:
Thanks @Scimon

PR #830, opened 2019-10-22
challenge-031 andrezgz solution

andrezgz:
Hi @manwar, I was very close to solve both tasks with eval, but changed my mind for the second one just before commiting


manwar:
I always look forward to your out of the box solutions. :-)
Thanks @andrezgz

PR #839, opened 2019-10-26
Solutions to challenge 031 problem 1 and 2 in Perl 6

noudald:
Another interesting challenge. Instead of using the obvious try/catch, I solved problem 1 in an unusual way. I compute the division in the stereographic projection. In the stereographic projection it is possible to divide by zero, which gives the north pole as result. Equivalent to checking if the denominator is zero, is to check if the division is equal to the north pole (the complex number i) in the stereographic projection.


noudald:
Keep the challenges coming!

manwar:
Just another gem from you.
Thanks for sharing with us, @noudald

PR #844, opened 2019-10-27
Snake solution 1

kolcon:


manwar:
This is really cool, since I can read Python, it doesn't look difficult to code it.
Great work, @kolcon

kolcon:
I'm also surprised I can write basically the same code...

PR #854, opened 2019-10-28
Challenge 30 by Jaldhar H. Vyas: typo fix.

jaldhar:
Doh! Laurent correctly points out my Perl6 solution for task 1 should compare the day of week with 7 not 0 as in Perl5.


manwar:
Thanks @jaldhar

PR #858, opened 2019-10-28
Added a test incase the widt check fails

Scimon:
Small fix as the thing to look for the available width my not work.


manwar:
Thanks @Scimon

PR #866, opened 2019-10-28
Solve 032 (ASCII Bar chart + Count instances)

choroba:
Plus a standard, many years old bash script to count instances.


manwar:
Thanks @choroba

PR #868, opened 2019-10-29
challenge-032 andrezgz solution

andrezgz:
Hi @manwar, extra credit and  bonus for the additional options on each task


manwar:
Thanks @andrezgz

PR #885, opened 2019-11-03
my solutions for challenge 32.. done fast and dirty

dcw803:
I like these questions, histograms of hashrefs..


manwar:
Thanks @dcw803

PR #892, opened 2019-11-04
Update Challenge 2 to specify the size of chart to draw

Scimon:
Updated the challenge 2 so it'll do any size of table.


manwar:
This is nice improvement :-)
Thanks @Scimon

PR #893, opened 2019-11-04
PWC 033

bwva:
This is my first submission to the Perl Weekly Challenge! Many thanks to @manwar and the Perl Community!


manwar:
Welcome on board and thanks for your contribution. @bwva

PR #927, opened 2019-11-11
Various slices both Sequences, Array and Hash

Scimon:


Scimon:
Challenge 2 included too.

manwar:
Thanks @Scimon as always for your encouragement and support.

PR #935, opened 2019-11-16
Challenge 034

andrezgz:
Hi @manwar, great topics for this week tasks. I've read a lot about dispatch tables in High Order Perl book for the second one.


manwar:
Thanks @andrezgz

PR #937, opened 2019-11-16
Solution to challenge 34 problem 1 and 2 in perl 6 by Noud

noudald:
Two short but interesting problems. Thank you!


manwar:
Thanks @noudald

PR #946, opened 2019-11-19
Here we go

Scimon:
Bit late as I felt crappy yesterday.
I left some comments so me crappy string manipulations would make sense.


manwar:
Thanks @Scimon

PR #948, opened 2019-11-19
Add missing 0 to word gap

mienaikage:
If @Firedrake does not mind, I've proposed a change to update their solutions. I was using their script to test my own, and noticed it had 6 zeros for the word gap rather than 7.


Firedrake:
Fine with me - indeed I read it as six zeros.

PR #963, opened 2019-11-24
ruben/drclaw solution for W35 ch1 and ch2 perl and raku

drclaw1394:
Combined ch-1 and ch-2 solutions in a single file for both perl and raku

The decoding avoids confusion with the extra bits at the end required to make up a complete byte.
Only a-z and space characters are processed.
This challenge was a little tricky. Too many 1's and 0's!


manwar:
Thanks @drclaw1394

PR #966, opened 2019-11-24
Solution for PWC #35 from Burkhard Nickels

burkhard-nickels:
Solution for PWC #35 from Burkhard Nickels


manwar:
Thanks @burkhard-nickels

PR #975, opened 2019-11-27
challenge-036 andrezgz solution

andrezgz:
Hi @manwar, the VIN validation task was hard cause the definitions about it are somewhat imprecise on wikipedia. The Knapsack Problem was great and I've been able to find a solution, checking all the possible combinations due to the small amount of input.


manwar:
Thanks @andrezgz

PR #978, opened 2019-11-28
challenge-036 andrezgz task 2 minor fixes

andrezgz:
It's me again @manwar, I've added some comments and fix others to the second task solution


manwar:
Thanks @andrezgz

PR #982, opened 2019-11-28
Uploading my first challenge attempt Challenge 36

saiftynet:
Not sure if I am doing it right...


manwar:
Welcome on board and thanks for sharing your first contributions @saiftynet

PR #990, opened 2019-12-01
imported my solutions to this week's challenges

dcw803:
Task#1 (VIN validation) is horribly detailed, although the checksum algorithm for US and Canada is interesting, and Task#2 (knapsack packing) seems so simple that it doesn't really need a program?


manwar:
Thanks @dcw803

PR #991, opened 2019-12-01
Solution for PWC #36 from Burkhard Nickels

burkhard-nickels:


manwar:
Thanks @burkhard-nickels

PR #999, opened 2019-12-03
challenge-037 solutions

andemark:


manwar:
You have been prolific contributor.
Thanks @andemark

PR #1003, opened 2019-12-04
challenge-037 andrezgz solution

andrezgz:
Hi @manwar. Hacking the second task I've learnt the odd behavior of the modulo operator with negative numbers. Check it out!


manwar:
Thanks for sharing knowledge as always. @andrezgz

PR #1008, opened 2019-12-07
Add Raku solution for 037-2

mienaikage:


mienaikage:
I'm not sure I understand the expected result of this challenge correctly

manwar:
Thanks @mienaikage

PR #1009, opened 2019-12-07
imported my solutions to challenge037

dcw803:
including 3 versions of ch-1 (the first challenge): 2 in Perl (one using a few Date::Manip functions, the other implementing those functions ourselves, i.e. doing leap year calculations, calculating number of days in each month, and finding out what day of the week a given date is by starting on a known data/day-of-week pair and rolling forward to the date you're interested in).

The 3rd version is a translation of the second Perl version into Postscript.  Yes, POSTSCRIPT, the language mainly used inside printers for page layout, is actually a full blown programming language.  It's a relatively straightforward translation of the Perl code into Postscript, but before that you need approx 3 pages of support routines - functions to append strings (!), and to produce a series of debugging messages "down the page".  I'm going for the "most unusual language to program one of these challenges in" prize, if indeed there is a prize:-)

### DIFF
--- a/challenge-037/duncan-c-white/README
+++ b/challenge-037/duncan-c-white/README
@@ -1,27 +1,59 @@
-Challenge 1: "Write a program to validate a given Vehicle Identification Number
-  (VIN). For more information, please checkout
-  https://en.wikipedia.org/wiki/Vehicle_identification_number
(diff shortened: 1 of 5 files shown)


manwar:
Thanks @dcw803

PR #1022, opened 2019-12-10
challenge-038 andrezgz solution

andrezgz:
Hi @manwar, this week solutions were straightforward. Regex for Task#1 and code reused from one of the first challenges for the other.


manwar:
Thanks @andrezgz

PR #1032, opened 2019-12-13
Solutions to challenge 38 problem 1 and 2 by Noud in Raku

noudald:
Hello Sajid,

The pre-christmas stress at work seems to get lower. That means I have a bit more time to work out the latest perl weekly challenges. ;) Again two interesting problems. I like problems where I can use grammar to solve them (problem 1). Thank you!

Best,
Noud


manwar:
Thanks @noudald

PR #1038, opened 2019-12-15
Finally added solutions to Week 1!

drclaw1394:
ch-1 and ch-2. perl and raku


manwar:
What a pleasant surprise. Happy to see you doing previous challenges.
Thanks @drclaw1394

PR #1046, opened 2019-12-17
drclaw/ruben solutions to w39 ch-1 and ch-2. perl and raku

drclaw1394:
Also noticed a missing words_alpha file from w38. This is added to repo now.


manwar:
Thanks @drclaw1394

PR #1047, opened 2019-12-17
Solutions for challenges 001-005

rjt-pl:


manwar:
Good job, thanks @rjt-pl

PR #1049, opened 2019-12-18
challenge-039 andrezgz solution

andrezgz:
Hi @mawar, it was nice to see my proposal as a task! Thanks a lot
The first one made me battle for a while because I considered noncontiguous periods of time registered on the guest book.


manwar:
Thanks @andrezgz

PR #1052, opened 2019-12-19
Remove unneeded pipes

mienaikage:
Cleanup for older solution, as I misunderstood captures in signatures


manwar:
Thanks @mienaikage

PR #1056, opened 2019-12-22
Task-1: Handles non-contiguous data Task-2: Offers 50+ RPN operations

saiftynet:
The data in **Task 1** has only overlapping times which merge into on cotinuous block.  The logs are also conveniently sequential.   The response therefore includes modifications to the data to allow handling of non-continuous data (periods with no residents) and out-of-sequence data entries.
**Task 2** response has been extended to convert many of Perl's internal operators, including string handling. Absent operations (e.g. asin etc) are also added, some coded as RPN scripts.  Stack manipulation operations (del, dup, swap)  are also included. Over 50 operations are included in roughly 100 lines of code.
[Blog regarding these submissions](http://blogs.perl.org/users/saif/2019/12/the-times-you-want-to-avoid-the-short-cuts.html)

### DIFF
--- a/challenge-039/saiftynet/perl5/ch-1.pl
+++ b/challenge-039/saiftynet/perl5/ch-1.pl
@@ -1,0 +1,69 @@
+#!/usr/env perl
+# Perl Challenge 39 Task 1
+# hotel lights.  Code detects non-contiguous stays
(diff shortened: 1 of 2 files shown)


manwar:
Thanks @saiftynet

PR #1059, opened 2019-12-23
ran out of time to get a Postscript code generator working..

dcw803:
(to show RPN expression order), so here's just the part 1 solution..


manwar:
Thanks @dcw803

PR #1060, opened 2019-12-23
updated solution for challenge 039

adamcrussell:
I noticed I had some bugs! must've happened in my hurry to get the solutions in on time. This seems to fix the issues.


manwar:
Thanks @adamcrussell

PR #1061, opened 2019-12-23
challenge-038 andrezgz ch-1 regex fix

andrezgz:
Quick regex fix after reading Kian-Meng Ang's review
https://perlweeklychallenge.org/blog/review-challenge-038/


manwar:
Brilliant !!!
Thanks @andrezgz

PR #1067, opened 2019-12-24
saiftynet's responses for the "Happy Lights" Challenges

saiftynet:
nice and staright forward for Christmas


manwar:
Thanks @saiftynet

PR #1071, opened 2019-12-25
Add solution to 039 by E. Choroba (Guest House Lights + RPN)

choroba:
Guest House Lights use Set::IntSpan which I learned from RFC: The Lazy Manager's Calendar with Inversion Lists: http://perlmonks.org/index.pl?node_id=908453


manwar:
Thanks @choroba

PR #1079, opened 2019-12-29
imported my (dcw803) solutions

dcw803:
Very simple tasks this week, fortunate as I left them to the last few hours:-)


manwar:
Thanks @dcw803

PR #1080, opened 2019-12-29
Just tried the same in R.

wanderdoc:
Dear Anwar,
I hope I did not break the rules to use the choroba's idea in R solution (ch-2). Otherwise please discard it!
Yours,
wanderdoc


manwar:
You certainly didn't break any rule. In fact you are encourage to share ideas.
Thanks for your contributions @wanderdoc.

Please drop us one line email to perlweeklychallenge@yahoo.com so that we can send you weekly challenge every Monday like other team members.

PR #1085, opened 2019-12-30
Branch for challenge 041

waltman:


manwar:
Welcome back after the short  break and thanks for sharing your solutions. @waltman

PR #1086, opened 2019-12-30
Solution to challenge #41. Thanks for the challenges this year!

Firedrake:


manwar:
Thanks for your support and encouragement. I hope we can continue this in the next year. @Firedrake

PR #1090, opened 2019-12-31
challenge-041 andrezgz solution

andrezgz:
Hi @manwar, both challenges were very similar! So I've solved really quickly this time!


manwar:
Thanks @andrezgz

PR #1093, opened 2019-12-31
challenge-023 ch-2 minor issue

andrezgz:
Found an issue when checking if 1 or lower numbers were prime and its prime factors


manwar:
I like how you keep your solutions crispy :-)
Thanks @andrezgz

PR #1105, opened 2020-01-04
Stashing 'n caching in hashes for smashing performance

saiftynet:
Trying multiple methods of improving performance.  The tasks make only minor demands on the code performance, but with  extra demands ( e.g. looking for 200 Leonardo numbers or 1,000,000 factorisations) simple code becomes too slow.  These solutions each try three ways  to improve performance.


manwar:
I love your experiments with the weekly challenge.
Thanks @saiftynet

PR #1129, opened 2020-01-12
Challenge 42 by Jaldhar H. Vyas

jaldhar:
It's been a difficult past two months but I'm back.  Now I'll start working on all the ones I've missed.


manwar:
We missed you. Glad you are back.
Thanks @jaldhar

PR #1135, opened 2020-01-13
Add solutions to 043 by E. Choroba

choroba:
There are two solutions for task 1: one uses combinatorics, the second
one is much simpler and faster.

The function in task 2 returns the number in its base if only one
parameter is specified, or in base 10 otherwise.


manwar:
Simply the best, thanks @choroba

PR #1140, opened 2020-01-16
challenge-043 andrezgz solution

andrezgz:
Hi @manwar, I'm back on track. Discovering self-descriptive number for higher bases was tough. The formula from Wikipedia works like a charm :)


manwar:
Nice to hear that, thanks @andrezgz

PR #1150, opened 2020-01-20
Challenge 43 by Jaldhar H. Vyas

jaldhar:
A little late but here you go.  Still working on the ones I missed.


manwar:
Thanks @jaldhar

PR #1151, opened 2020-01-20
Challenge number 1. Functional Raku for the win

Scimon:
I hope things at home are going OK and your Dad is good.

No rush on this but I do think it's a nice solution. I may even blog about it.


manwar:
Thanks @Scimon

PR #1154, opened 2020-01-20
Updated solution W43 ch-2.p6 to use Raku built in base handling

drclaw1394:
Just needed to look a little deeper in the docs! Removes a bunch of lines


manwar:
Thanks @drclaw1394

PR #1157, opened 2020-01-21
Helps if I use the correct final number.

Scimon:
Sorry Mohammed, for some reason (probably left over from the first challenge) I went with 100 not 200. Fixed it.


manwar:
Thanks @Scimon

PR #1163, opened 2020-01-23
challenge-044 andrezgz solution

andrezgz:
Hi @manwar. This week tasks were really challenging to me. I think I've proposed a decent solution for each one. Looking forward to compare with others!


manwar:
Thanks @andrezgz

PR #1164, opened 2020-01-24
Blending Numbers and operations together

saiftynet:
Hope your dad is getting better mate...


manwar:
Dad is doing well and is back home now.
Thanks @saiftynet

PR #1167, opened 2020-01-25
Add a link to E. Choroba's blog about 044

choroba:
Also, update the solutions: don't skip the last mask in task 1 and
don't use an additional variable for a negligible reduction of the hash.


manwar:
Thanks @choroba

PR #1176, opened 2020-01-29
Quick attempts at challenge #45

davorg:


manwar:
Welcome back and thanks for your contributions @davorg

PR #1180, opened 2020-01-29
Another one-liner

pjscott:
Another one-liner


manwar:
I simply love your one-liner, thanks @pjscott

PR #1182, opened 2020-01-29
Add 045 solution (Square Secret Code + Source Dumper) by E. Choroba

choroba:
Quine added as an alternative solution


manwar:
Thanks @choroba

PR #1187, opened 2020-01-31
challenge-043 task-1 fix

andrezgz:
Hi @manwar I've removed an unnecessary line noted by Ryan Thompson in his review. As he said:
"Andrezgz defines, but does not even need, a hash of the available numbers, 1, 2, 3, 4, and 6."


manwar:
Thanks @andrezgz

PR #1188, opened 2020-01-31
challenge-045 andrezgz solution

andrezgz:
Hi @manwar, I've made a quine for the second task. I'm proud of such an ugly code :)


manwar:
Thanks @andrezgz

PR #1191, opened 2020-02-01
Perl solution to Task #1 ("Square Secret Code") of the Perl Weekly Ch…

PerlMonk-Athanasius:
…allenge #045

On branch branch-for-challenge-045
Changes to be committed:
	new file:   challenge-045/athanasius/perl/ch-1.pl


manwar:
Thanks @PerlMonk-Athanasius

PR #1199, opened 2020-02-02
imported my solutions

dcw803:
two nice easy but interesting challenges this week.. did decoding as well as encoding in the first task


manwar:
Thanks @dcw803

PR #1211, opened 2020-02-03
I like one-liners...

pjscott:
I like one-liners, what can I say...


manwar:
Thanks for sharing your knowledge @pjscott

PR #1218, opened 2020-02-06
imported my solutions …

dcw803:
…including the second one in postscript to get on the guest page again:-)


manwar:
Thanks @dcw803

PR #1225, opened 2020-02-08
Challenge 046 noud

noudald:
I liked task #2 so much I wrote a short blog post about it, see blog.txt. A very well picked challenge! It took me quite a while before I figured out how to do it. Thanks!


manwar:
Thanks @noudald

PR #1234, opened 2020-02-10
Morning

Scimon:
I wasn't sure if it was OK to move to .raku extensions... so I did both. :)

Feel free to delete one set.


manwar:
Good morning, I am not too sure about .raku extension. I will keep it .p6 for the time being. Also parser only understands .p6 for now. Thanks for your support @Scimon

PR #1240, opened 2020-02-12
challenge-047 andrezgz solution

andrezgz:
Hi @manwar, I've reused code from previous challenges for the first task. I was close to a one-liner for the second one.


manwar:
Thanks @andrezgz

PR #1241, opened 2020-02-12
Challenge 47

davorg:


manwar:
It is always an honour to have your contributions to the weekly challenge, thanks @davorg

PR #1244, opened 2020-02-12
Solve 047 (Roman Calculator and Gapful Number) by E. Choroba

choroba:
Code to handle Roman Numberals extracted from 010.


manwar:
Master of code-reuse ;-)
Thanks @choroba

PR #1253, opened 2020-02-15
please find my solutions - both in perl, ch-2 in postscript as well

dcw803:
I like getting into the "Guest" section too:-)


manwar:
Thanks @dcw803

PR #1259, opened 2020-02-16
Perl solutions to Tasks 1 & 2, and Raku solution to Task 2, of Challe…

PerlMonk-Athanasius:
…nge 047.


manwar:
Thanks @PerlMonk-Athanasius

PR #1264, opened 2020-02-16
Challenge 47 by Jaldhar H. Vyas - problem 2

jaldhar:
My solutions for problem 1 have a couple of bugs so they will be submitted a little late.


manwar:
Thanks @jaldhar

PR #1266, opened 2020-02-17
Challenge #48

davorg:


manwar:
You surprised me with the fastest fingers first ;-)
Thanks as always, @davorg

PR #1274, opened 2020-02-17
Challenge 47 by Jaldhar H. Vyas - problem 1 and blog.

jaldhar:
So here are the other solutions for challenge 47.  I implemented actual addition and subtraction of Roman numerals which I don't think anyone did.  Do check it out and the accompanying blog post.


manwar:
Thanks @jaldhar

PR #1276, opened 2020-02-18
Smarter implementation of task 2.

fluca1978:
I realized that task 2 could have been done with a simpler approach.
I've already blogged about this approach, so it is fine to keep the old solution I already submitted (and merged).


manwar:
Thanks @fluca1978

PR #1277, opened 2020-02-18
Solutions for challenge #48.

Firedrake:
ch-2a.pl is an implementation of ch-2.p6 to see if perl6 really is that slow. It is.


manwar:
Thanks @Firedrake

PR #1281, opened 2020-02-19
challenge-048 andrezgz solution

andrezgz:
Hi @manwar , both tasks were great for improving the solution after the first try


manwar:
Thanks @andrezgz

PR #1282, opened 2020-02-20
It's 50 swords not 500

holli-holzer:


manwar:
Good catch !!!
Thanks @holli-holzer

PR #1286, opened 2020-02-20
My submission for challenge 48

phillip-harris:
My submission for challenge 48


manwar:
Thanks @phillip-harris

PR #1289, opened 2020-02-21
xkr47 / 048 / Rust version!

xkr47:
I already participated in Raku language but thought I'd challenge myself a bit as I'm learning the Rust language as well.. So if you can consider accepting Rust as a language, here's my shot at it. I chose to implement it very similarly to my Raku version to make them comparable. If you can't, I guess interested people can read this PR instead :)

### DIFF
--- a/challenge-048/xkr47/rust/Cargo.toml
+++ b/challenge-048/xkr47/rust/Cargo.toml
@@ -1,0 +1,16 @@
+[package]
+name = "pwc-048"
+version = "0.1.0"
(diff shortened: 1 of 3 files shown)


manwar:
I am very impressed by your positive energy. It is always welcome in any language you feel comfortable in. I am not sure if you have read the below blog:

https://perlweeklychallenge.org/blog/guest-contribution/

If you noticed, we have had contributions in many different languages. 

So thank you for your contributions as always @xkr47

PR #1298, opened 2020-02-23
Ruben's solutions for W48 ch-1 and ch-2. perl and raku

drclaw1394:
ch-2.raku takes a while to run for some reason.


manwar:
Thanks @drclaw1394

PR #1303, opened 2020-02-24
Challenge 49

Scimon:
This is not Thread safe though it would be simple enough to make it so using OO::Monitors on the methods.


manwar:
Thanks @Scimon

PR #1310, opened 2020-02-25
added impossibility check

holli-holzer:


manwar:
Thanks @holli-holzer

xkr47:
Hi,
Out of interest I had to try your version, but found some funnies:
* for 995 your version reports 1111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111110 but I get 1110000110 (= 1115578 * 995)
* for 98 your version reports `Impossible` but I get `11000010` (= 112245 * 98)
All my numbers are in base 10.

holli-holzer:
Yes, I just realised it's buggy.

Try 
```
say (1 .. Inf).map( *.base(2).Str ).first( * %% $n );
```

xkr47:
It works for my tests.. and it's faaast! :)

xkr47:
btw, `base(2)` already returns `Str`

holli-holzer:
That's true. The optimizer will magicwand that away. 
I figured I leave it in to make it clear we have a string of 1s and 0s.

PR #1318, opened 2020-02-27
ch-1

jacoby:
using sprintf '%b'

Dunno if I'll do ch-2, but this is in.


manwar:
We are happy you contributed one at least. Thanks @jacoby

PR #1324, opened 2020-02-28
Add alternative solution to LRU Cache using a linked list

choroba:
Also, remove an unused subroutine from the original solution. For
larger caches, the linked list solution is much faster (benchmark will
be shown in a blog).


manwar:
Coolest solution. Thanks @choroba

PR #1358, opened 2020-03-04
Use enum instead of bitmask literals, remove the nested hash

choroba:
Singles can be represented by another bitmask value if we ignore it
when checking for left and right boundaries.


manwar:
I am going to rely on your blog for explanation. Thanks @choroba

PR #1360, opened 2020-03-05
Add blogpost from E. Choroba(050: Merge Intervals + Noble Integer)

choroba:
Also, remove the useless condition from the code.


manwar:
Thanks @choroba

PR #1382, opened 2020-03-09
Challenge-051 solutions by phillip-harris

phillip-harris:
This was challenge was fun, thanks.


manwar:
Thanks @phillip-harris

PR #1384, opened 2020-03-09
Add solutions to 051 by E. Choroba: 3 Sum + Colourful Number

choroba:


manwar:
Great solutions, specially the second task. Thanks @choroba

PR #1388, opened 2020-03-09
better semantics

holli-holzer:


manwar:
All these changes going above my head ;-)
Thanks @holli-holzer

holli-holzer:
Well, this is the core:

```
@digits
		.pairs
		.combinations( 1..* )
		.grep({ .&keys.&are-consecutive })
		.map({ .&values })
	;
````

I wanted to support arbitrary values, not just from 100 to 999. 
So the idea is to split each value into digits and turn them into index/value pairs. 
Then let raku figure out the combinations of the pairs. But then you have non consecutive pairs in the list. So i filter those out. And finally I map back from the pair to the original values.
And then later on we take the product of the values and see if they are uniq.

 Simple, really.

holli-holzer:
I mean, I used speaking function and variable names and everything. Try reading it again. If you have trouble understanding a specific part, please ask.

PR #1392, opened 2020-03-10
challenge-050 andrezgz solution

andrezgz:
Hi @manwar, I recovered the lost time. Thanks in advance for your post about the solutions for these week tasks. I've found your test cases (borrowed from E. Choroba) well-suited for making a robust code. I added them to mine.


manwar:
It is team work, great. Thanks @andrezgz

PR #1393, opened 2020-03-10
challenge-039 andrezgz task1 fix

andrezgz:
Hi @manwar, when I was coding the solution for the first task of challenge 50 (Merge Intervals), I reminded doing something similar some time ago. So I grep my older solutions and I've found the guest house task in challenge 39. A little omission appeared in a condition while I was reviewing it, but now is fixed.


manwar:
Thanks @andrezgz

PR #1395, opened 2020-03-10
Challenge-051 solutions by saiftynet

saiftynet:


manwar:
Didn't you use EZPWC to submit this PR?
Thanks @saiftynet

saiftynet:
I did, but the new versions don't declare it, because I guess others may
not be keen to move the emphasis from their code to a tool.

best wishes

saif

PR #1409, opened 2020-03-15
Add blog post by E. Choroba about 051 (3Sum and Colourful Numbers)

choroba:
Also, simplify the solution of the second task: we don't need to pad
the array.


manwar:
Thanks @choroba

PR #1425, opened 2020-03-18
s/https/http/g

rjt-pl:
You even told me the links were broken, and I thought it was only because I had set them to auto-publish. It turns out there were two issues. I didn't even notice I'd typed `https://` instead of `http://`.


manwar:
I didn't notice that myself. Thanks @rjt-pl

PR #1429, opened 2020-03-20
Add solutions to 052 (Stepping Numbers + Lucky Winner) by E. Choroba

choroba:
Stepping Numbers are optimized, try increasing the length to 16.

Lucky Winner is implemented as an interactive game. The starting
player can always win unless they make a mistake, so the computer
always wins if it starts; if you start, it looses unless you make a
mistake.


manwar:
Thanks @choroba

PR #1434, opened 2020-03-21
Solution for challenge 052 task 1 and 2 in Raku by Noud

noudald:
Interesting challenges this week! Especially task 2. When you figure out that it is a min-max game, the solution becomes elegant. Looking forward to the challenges of next week.


manwar:
Thanks @noudald

PR #1435, opened 2020-03-21
challenge-052 andrezgz solution

andrezgz:
Hi @manwar, just today I've found time to do the tasks. There's a lot of movement these days, and the this challenge is a way to escape from all of that


manwar:
Thanks @andrezgz

PR #1437, opened 2020-03-21
Add link to a blog post by E. Choroba

choroba:
Also, fix the Lucky Winner according to it: init_arg was 0 instead of
undef and turns were implemented twice.


manwar:
Thanks @choroba

PR #1448, opened 2020-03-22
imported my solutions for challenge 52..

dcw803:
task 1 was simple, task 2 rather badly defined, if I've understood it right: first player always wins, both players try to ensure that they can take the £2 coin


manwar:
Thanks @dcw803

PR #1453, opened 2020-03-23
Pwc53

fluca1978:
Quick and dirty solutions in raku.


manwar:
I wonder how you manage to get all done so quickly.
Thanks @fluca1978

fluca1978:
The quality of my raku solutions is decreasing, since I've no the right mindset in this covid-19 period. I see there is a lot of other people who solves the puzzles before me, so they are quicker than me!

manwar:
> The quality of my raku solutions is decreasing, since I've no the right mindset in this covid-19 period. I see there is a lot of other people who solves the puzzles before me, so they are quicker than me!

I don't blame you.
We are all on the same boat.

PR #1458, opened 2020-03-23
mostly the vowel one

jacoby:
But also in Node and Rust!


jacoby:
changed filenames because I reversed the tasks in my head.

manwar:
Thanks @jacoby

PR #1463, opened 2020-03-24
challenge-053 andrezgz solution

andrezgz:
Hi @manwar, we're on quarantine & holiday so I did a session of online pair programming with a friend to solve this challenge


manwar:
That is the best use of quarantine. Why don't share your experience in a short blog.
Thanks @andrezgz

andrezgz:
I'll try my best but I can't promise anything cause I'm working long hours these days

PR #1487, opened 2020-03-29
imported this week's challenges; nearly forgot to do them..

dcw803:
task 1 was simple so I livened it up slightly via a Matrix class (inline Perl package); task 2 was vaguely interesting, as a sort of recursive generator


manwar:
Thanks @dcw803

PR #1490, opened 2020-03-30
Here we go

Scimon:
I may look at Extra Credits but I think I need to completely rewrite it to use caching


manwar:
Thanks @Scimon

PR #1501, opened 2020-04-01
edges -> next (fix silly search/replace error)

rjt-pl:
Found this during my review. When I was blogging, I wanted to rename a variable `%next` to `%edges`. Search/replace without testing was obviously not smart.


manwar:
Thanks @rjt-pl

PR #1502, opened 2020-04-01
Add 054 by E. Choroba (K-th Permutation and Collatz Conjencture)

choroba:
Permutation is solved recursively character by character.

To solve the Extra Credit task of the Collatz Conjencture, I first
used a heap, but in the end solving it with a pure array was much
simpler and even faster (and not terribly memory hungry either).


manwar:
Thanks @choroba

PR #1506, opened 2020-04-03
So here's a thing

Scimon:
My attempt at the extra credits.... it takes a while.

I tried caching the chains but that was eating memory so I decided to go with parallel tasks instead.


manwar:
Thanks @Scimon

PR #1523, opened 2020-04-05
imported my solutions to challenge-054

dcw803:
I always hate permutation questions, but I'm quite pleased to have come up with a direct generation of the Kth permutation algorithm (w/o calculating all previous permutations).  The second question was quite fun, especially the "longest 20 sequences" extra credit part (note the "longest 20 sequences" output is included (file "ch-2:-100000.output")


manwar:
Thanks @dcw803

PR #1543, opened 2020-04-10
challenge-055 andrezgz solution

andrezgz:
Hi @manwar, thanks for keeping this proyect alive!


manwar:
Thanks for your support and encouragement @andrezgz

PR #1546, opened 2020-04-10
Ruby solutions for ch#55

Doomtrain14:


manwar:
Another language added to the list, thanks @Doomtrain14

PR #1553, opened 2020-04-12
Solutions for challenge 055 task 1 and 2 in Raku by Noud

noudald:
Both tasks were more difficult than I thought they would be. I'm curious to find out if there were smart (i.e. shorter) solutions to these tasks. Looking forward to the tasks of next week!


manwar:
Thanks @noudald

PR #1561, opened 2020-04-12

dcw803:
Two nice questions this week, for some reason I found the second task quite hard to get my head round in order to start thinking about it, but easy after I'd started..


manwar:
Thanks @dcw803

PR #1562, opened 2020-04-12
update to ch-2.pl

jaredor:
Removed non-core debugging library.


manwar:
Welcome on board and thanks for your first contributions @jaredor

PR #1583, opened 2020-04-16
challenge-056 andrezgz solution

andrezgz:
Hi @manwar, cool tasks this week. The one with the tree was harder. Do you know a simple way to pass the tree nodes as arguments to the script?

### DIFF
--- a/challenge-056/andrezgz/perl/ch-1.pl
+++ b/challenge-056/andrezgz/perl/ch-1.pl
@@ -1,0 +1,42 @@
+#!/usr/bin/perl
+
+# https://perlweeklychallenge.org/blog/perl-weekly-challenge-056/
+# Task #1
+#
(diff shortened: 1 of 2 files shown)


manwar:
Thanks @andrezgz 

I defined the tree in the script instead of accepting from the command line.
https://github.com/manwar/perlweeklychallenge-club/blob/master/challenge-056/mohammad-anwar/perl/ch-2.pl

Some came up with the string like "5(4(11(7)(2)))(8(13)(9(1)))".
https://github.com/manwar/perlweeklychallenge-club/blob/master/challenge-056/simon-proctor/raku/ch-2.p6

andrezgz:
I defined the tree in the script too. I like Simon's approach though. Maybe I'll give it a try.

Thanks a lot!

PR #1584, opened 2020-04-17
I can't seem to stop poking at this

Scimon:


manwar:
I am unable to decode the magic. I might PM you later.
Thanks @Scimon

PR #1597, opened 2020-04-19
better scoping

holli-holzer:


manwar:
One day, I will revisit and learn from this. Thanks @holli-holzer

PR #1599, opened 2020-04-19
Added solution ch-2.pl

manfredi:


manwar:
Welcome on board and thanks for your first contribution @manfredi

PR #1600, opened 2020-04-19
imported my solutions for this week's tasks..

dcw803:
first one very simple; second one surprisingly complex (I had to build the machinery for representing a binary tree first, parsing it from a string form on the command line, before writing the path-sum-finding code)


manwar:
Thanks @dcw803

PR #1605, opened 2020-04-19
Add necessary check for required option -k

jaredor:
Thank you Mohammad, for catching the warning for null comparison against the user option for -k. I've implemented your suggestion, but I hope you don't mind that I recast it using my style: I like "die ... unless ..." for going through validation checks. Cheers, --Jared


manwar:
Thanks @jaredor

PR #1613, opened 2020-04-20
Add blog post by E. Choroba about 056 (Diff-K + Path Sum)

choroba:
Also, fix the order of indices in the solution to Task 1, and add
another solution that works for repeated numbers.


manwar:
Thanks @choroba

PR #1622, opened 2020-04-25
challenge-057 andrezgz solution

andrezgz:
Hi @manwar, dealing with the full binary tree was fun and took me a couple of hours to leave it the way I wanted


manwar:
Thanks @andrezgz

PR #1636, opened 2020-04-26
ruben's solutions for w56 and w57 ch-1 and ch-2. perl and raku

drclaw1394:
Playing catchup with last weeks challenge. Solutions added
No pretty print bonus for my tree inversion for this weeks ch-1 :(


manwar:
Thanks @drclaw1394

PR #1637, opened 2020-04-26
imported my solutions (including two attempts at task 1's bonus question…

dcw803:
1. pretty-printing the tree using GraphViz,
2. drawing proper ascii pretty-print of the tree - then revised that to include support for printing non-full binary trees, where missing branches are not drawn.


manwar:
Thanks @dcw803

PR #1643, opened 2020-04-28
Cheok Yin's submission for challenge #058 Task 2

E7-87-83:


E7-87-83:
haven't settled to output an error msg when the inversion requirement is impossible

PR #1648, opened 2020-04-28
Add solutions to 058: Compare Version & Ordered Linup by E. Choroba

choroba:
Using Perl::Version (as "version" gets the alpha releases wrong) and
List::UtilsBy.


manwar:
Thanks @choroba

PR #1650, opened 2020-04-29
Solution for task #2

jo-37:


manwar:
Welcome on board and thanks for your first contribution. @jo-37

PR #1662, opened 2020-05-03
imported my solutions to this week's problems; task 2 is very interesting

dcw803:
my solution works, but doesn't scale even to the 64-person example, clearly some better algorithm needed.


manwar:
Thanks @dcw803

PR #1688, opened 2020-05-09
Solution to challenge 059 task 1 and 2 in Raku by Noud

noudald:
I couldn't find the function of task 2 in the online encyclopedia of integer sequences (OEIS). This either means that I made a mistake in the algorithm, or that the sequence really isn't known in the OEIS. If the latter is the case I would like to understand more about this function. Perhaps it would be a good idea to add this function to the OEIS.


manwar:
Thanks @noudald

PR #1692, opened 2020-05-09
Solution updated.

sangeetkar:
- Merging Listy.pm and ch-1.pl to have one file for the 1st challenge
- Minor changes in ch-2.pl


manwar:
Thanks @sangeetkar

PR #1712, opened 2020-05-12
Add blog post by E. Choroba about 060: Excel Column + Find Numbers

choroba:
Also, fix the edge case where input numbers of the target length were
ignored in "Find Numbers".


manwar:
Thanks @choroba

PR #1725, opened 2020-05-17
plis-acm finish challenge 060 task #1 and #2

plis-acm:
My first challenge 😄 

Solution to task #2 maybe inefficient :(


manwar:
Welcome on board !!!
Thanks for your first contributions. @plis-acm

PR #1730, opened 2020-05-18
Solutions for challenge #61.

Firedrake:
Hope the use of "perl" and "raku" directories is convenient for you.


manwar:
Since Perl6 renamed to Raku, it doesn't make sense to call Perl as Perl5. However you use either when submitting your solutions. Thanks for being the first contributor this week. @Firedrake

PR #1736, opened 2020-05-18
Solve 061 by E. Choroba

choroba:
Unfortunately, using List::Util isn't possible for Product SubArray,
see https://rt.cpan.org/Ticket/Display.html?id=132653.


manwar:
Thanks @choroba

PR #1776, opened 2020-05-31
imported solution to first task, sorry no time to do second task

dcw803:
... and didn't much like the question anyway:-)


manwar:
Thanks @dcw803

PR #1795, opened 2020-06-07
Optimize LCS via Suffix Tree

choroba:
Also, make it general by not using high unicode characters as
separators, but rather using arrays instead of strings to store the
characters, and making the markers a special array elements.

Moreover, optimize the coverage computing by iterating over the
numbers rather than the covered characters.


manwar:
Thanks @choroba

PR #1796, opened 2020-06-07
Solve 062/1 Sort Email Address by E. Choroba

choroba:
Use Email::Address::XS to correctly parse addresses containing @ and "
in the user part.


manwar:
Thanks @choroba

PR #1801, opened 2020-06-08
Pwc64

fluca1978:
My solutions in raku. The first task is not exhaustitive, maybe I will find a better solution later on.


manwar:
Thanks @fluca1978

PR #1829, opened 2020-06-15
Challenge 64 fix

brtastic:
I've noticed I misunderstood the last challenge #1 and only returned the path sum. This is a small tweak to return the path as well


manwar:
Thanks @brtastic

PR #1842, opened 2020-06-21
Cheok Yin's submission

E7-87-83:
It's left to blog in or before the tomorrow morning (GMT+8.00 timezone)

 :)


manwar:
Thanks @E7-87-83

PR #1843, opened 2020-06-21
Solutions to challenge 65 task 1 and 2 in Raku by Noud

noudald:
I didn't understand task 2 well. I gave some alternative solutions how I understood the exercise.


manwar:
Thanks @noudald

PR #1847, opened 2020-06-21

dcw803:
the first task was nice and simple, but I'm afraid the second task doesn't clearly define what a partition is.  See my README for a discussion of that.  I can't be expected to debug the specification and then solve that, so I haven't got the desired answers eg. for example 1 I get four answers not 3.  If "aa,baab" is an answer, why isn't "aa,aa" also an answer?  So I gave up.  Better luck next time, Mohammed - but keep up the (mostly) good work.


manwar:
Thanks @dcw803

PR #1848, opened 2020-06-21
Add solutions to 065 by E. Choroba

choroba:
There are two solutions for 065/1 Digits Sum. The first one is the
naive one which uses the brute force. The second solution generates
only the possible numbers and seems to be much faster: on my machine,
it runs 3 times faster for length 7 and sum 42.

The Palindrome Partition adds one more solution to the "aabaab" case:
["aa", "aa"]. It's not clear why this solution is invalid (Must the
solutions be adjacent? Should each solution be never part of another
solution?) Fixing the algorithm might not be straight forward, as
additional constraints might need information that's not available in
the current implementation.

### DIFF
--- a/challenge-065/e-choroba/perl/ch-1.pl
+++ b/challenge-065/e-choroba/perl/ch-1.pl
@@ -1,0 +1,16 @@
+#!/usr/bin/perl
+use warnings;
+use strict;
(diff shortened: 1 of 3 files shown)


manwar:
Thanks @choroba

PR #1852, opened 2020-06-22
Pwc66

fluca1978:
Brutal raku solutions.
Please note that the directory `challenge-066` was missing when I started.


manwar:
Thanks @fluca1978 
Just FYI, I have updated one of the example of Task #1 as it was mathematically incorrect.

PR #1882, opened 2020-06-29
Had last week off but I'm back.

Scimon:
I had taken the week off for the conference but didn't feel up for it so just chilled out. 

Here's this week's challenge.


manwar:
Hope you had great time. thanks @Scimon

PR #1886, opened 2020-06-29
Still an enhanced task 1.

fluca1978:
Sorry to bother, but here is a more compact and correct version of task 1 in raku


manwar:
Thanks @fluca1978

PR #1896, opened 2020-07-02
rework 058-2

jo-37:


manwar:
Nice to see, you still care about your past solutions, thanks @jo-37

PR #1903, opened 2020-07-04
python 3 code for challenge 67

waltman:
These turned out to be a lot simpler than my Perl versions, in large part
because you don't have to worry about references in Python.


manwar:
Thanks @waltman

PR #1919, opened 2020-07-07
python code for challenge 68

waltman:
I don't know why I started writing Python versions of these challenges but here you go!


manwar:
Thanks @waltman

PR #1935, opened 2020-07-12
add raku solution task1 task2

jeongoon:
May be too late to request. Thanks anyway.


manwar:
It is never too late, you are well before the deadline :-)
Thanks @jeongoon

PR #1937, opened 2020-07-13
Add ch-1.pl from strobogrammatic numbers task

simonminer:
Includes solution to task 1 on strobogrammatic numbers.


manwar:
Thanks @simonminer

PR #1947, opened 2020-07-15
ash: 069-2 using Raku’s sequences(!)

ash:


manwar:
Too much fun with Raku :-) @ash

PR #1949, opened 2020-07-17
Branch for challenge 069

waltman:
Added python code for the 2 tasks.
Sped up the perl code by using tr instead of iterating over the string.


manwar:
Thanks @waltman

PR #1950, opened 2020-07-17
Simon's Challenge 069 solutions

simongreen-net:


manwar:
Thanks @simongreen-net

PR #1968, opened 2020-07-21
Branch for challenge 070

waltman:
Added python solutions for challenge 70
Removed unneeded `scalar` from perl/ch-2.pl


manwar:
Thanks @waltman

PR #1978, opened 2020-07-25
Add solutions to 070 by E. Choroba

choroba:
Character Swapping - I don't know what's the problem with $C > $O. For
a simple fast implementation, I needed a different additional
constraint: $C + $O < $N.

Grey Code Sequence - Two solutions. The shorter iterative one is
slightly faster, but the difference decreases with the sequence
length.


manwar:
Thanks @choroba

PR #1983, opened 2020-07-26
imported this week's solutions; two easy problems

dcw803:
two nice easy questions this week, although there wasn't much elegance to my task 1 solution..

total solving time: 40 minutes.


manwar:
Thanks @dcw803

PR #1989, opened 2020-07-27
Pwc71

fluca1978:
My solutions in raku.

Please note that it is not clear, to me, what "neighbour" means in task 1, so I've implemented it assuming "neighbours" that is looking at the preceding and following elements.


manwar:
Thanks @fluca1978

PR #1993, opened 2020-07-28
Submit challenge 70 solutions

boblied:
First-time submission, working through github flow.  Last week's problem, but better late than never.


manwar:
I always encourage to share even if you miss the deadline. You don't want to waste your efforts.
Thanks and congratulation on your first contribution. @boblied

PR #2004, opened 2020-07-30
Task 2: this time with a SINGLY-linked list

jacoby:
When i wrote this at first, I used an existing list implementation hat was doubly-linked. 

I blame/credit Walt Mankowski for bringing this to my attention and forcing me to recode.


manwar:
Thanks @jacoby

PR #2007, opened 2020-08-01
for task 1, make the elements on the array unique

waltman:
I updated my code because I just realized the elements in task 1 needed to be unique.


manwar:
wow, first blog ever for us, thanks @waltman

PR #2016, opened 2020-08-02
miss out the uniqueness requirement in Task #2 !

E7-87-83:
The Lisp script is not fixed. The Perl script is fixed.


E7-87-83:
miss out the uniqueness requirement in Task #1 !

manwar:
Thanks @E7-87-83

PR #2019, opened 2020-08-03
Challenge 72

Scimon:
Here you go Mohammad, I hope you had a nice Eid.


manwar:
Thanks @Scimon 
I had nice Eid, kids enjoyed the day which is important :-)

PR #2025, opened 2020-08-03
Moved challenge answers into raku directory as appropriate

therealjcm:
I didn't fully read the instructions and initially thought that raku solutions stayed in my root challenge directory. I've moved my answers to the appropriate directory.


manwar:
I had moved it already :-)

PR #2028, opened 2020-08-03
Solutions for #72, "Range"

fecundf:
2 In Perl, 1 in Raku


manwar:
Welcome on board and thanks for your first contributions @fecundf

PR #2030, opened 2020-08-04
Solve 072 (Trailing Zeros and Lines Range) by E. Choroba

choroba:
In 072/1, ignore the N <= 10 condition. We can solve it fast for much
larger Ns, also, it's N == 25 where it starts to get tricky and
interesting.

In 072/2, specify the input filename as the last argument to make it
optional. $. is needed in the range operator as only literal numbers
are interpreted as line numbers directly.


manwar:
Thanks @choroba

PR #2032, opened 2020-08-04
Better messages on bad inputs

fecundf:
My arg handling in Raku needed improvement, the "where" clause should say what is wrong when rejecting input. Fixed by adding "die" statements.


manwar:
Thanks @fecundf

PR #2034, opened 2020-08-04
Challenge 72 Completed by Jeon

jeongoon:
I have more time this week so I have a look into some other language.


manwar:
You seemed to broke all the records, thanks @jeongoon

PR #2036, opened 2020-08-04
untaint arguments

jo-37:


manwar:
Nice attempt :-) @jo-37

PR #2044, opened 2020-08-06
Add solutions in Perl and Elm by Lance Wicks

lancew:
Hope this is ok. Perl is just fine.
Elm is I hope ok


manwar:
Welcome on board and thanks for your contributions @lancew

PR #2048, opened 2020-08-07
ch-072/jeongoon bug fixed in both perl and raku when matching "/0+$/" has no result

jeongoon:
thank you.


manwar:
Thanks @jeongoon

PR #2051, opened 2020-08-09
Simon's solution to challenge 72

simongreen-net:


manwar:
Thanks @simongreen-net

PR #2052, opened 2020-08-09
Solution to challenge 072 task 1 and 2 in Raku by Noud

noudald:
Looking forward to the challenge of next week!


manwar:
Thanks @noudald

PR #2053, opened 2020-08-09
[ch-072/jeongoon] add brute force method for haskell/ch-1.hs; add Haskell for  [ch-069:Task1]

jeongoon:
Haskell is mystery...


manwar:
Thanks @jeongoon

PR #2054, opened 2020-08-09
Cheok Yin's submission for Challenge #072

E7-87-83:
Besides the two Perl tasks, have a Lisp solution again! :)


manwar:
Thanks @E7-87-83

PR #2059, opened 2020-08-10
Challenge 073 solution, both tasks

shawnw:
My first submission to the weekly challenge.

Uses perl + sqlite (Depends on `DBI` and `DBD::SQLite` 1.59_01 or newer).


manwar:
Welcome on board and thanks for your first contributions @shawnw

PR #2065, opened 2020-08-10
Added Perl solutions by Pavel Kuptsov for Task-1

poizon:
Added solutions by Pavel Kuptsov for Task-1


manwar:
Welcome on board and thanks for your first contribution @poizon

PR #2067, opened 2020-08-10
Weekly challenge #73 raku solution by Jason Messer

therealjcm:
Weekly challenge #73 raku solution by Jason Messer


manwar:
Thanks @therealjcm

PR #2078, opened 2020-08-15
Pull Request for Challenge 73, Part 1

WmMonty:
I had some trouble with git because I'm trying to use a different machine, and git's just plain weird. I hope That I got everything right.

Crossing my fingers!


manwar:
Thanks @WmMonty

PR #2079, opened 2020-08-16
Challenge 73 solution by Simon

simongreen-net:


manwar:
Thanks @simongreen-net

PR #2081, opened 2020-08-16
Challenge 73 Task 2 only, 2 alternative implementations

xkr47:


manwar:
Thanks @xkr47

xkr47:
Thank you!

I realize now that I could even have gotten rid of the dependency of the external module by just handling "no previous value" as a special case instead of starting from the maximum of the type as the minimum value seen so far..

PR #2103, opened 2020-08-18
So... still don't get why it's FNR but this matches the spec

Scimon:


manwar:
I can understand why it is confusing many members.
The word "First" is misleading, we assume "first" is always from "top". In this case, it was "first" from bottom.
If we change the direction, "first" also changes. Therefore I updated the task description slightly to help others.
Thanks for sharing the solution @Scimon

PR #2108, opened 2020-08-19
[ch-074/jeongoon] Haskell, Common-lisp solution

jeongoon:
Thank you.
I'm wondering how earlier can I access the new challenge?
because my day off has been changed from weekend, so I feel like I have to access earlier.

### DIFF
--- a/challenge-074/jeongoon/common-lisp/ch-1.lsp
+++ b/challenge-074/jeongoon/common-lisp/ch-1.lsp
@@ -1,0 +1,89 @@
+;; tested with:
+;; sbcl --script ch-1.lsp 10
+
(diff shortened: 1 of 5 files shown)


manwar:
If you want to access the challenge earlier than Monday then you can join the "Early Bird Club". All members of the club gets the challenge email before Friday, sometimes even Monday/Tuesday. Just like this week, I sent the challenge on Tuesday evening. If you want to have access then please drop us a line at perlweeklychallenge@yahoo.com. .We would then send you the link to the next week challenge i.e. week 75 instantly.
Thanks for your support and contributions @jeongoon

PR #2109, opened 2020-08-19
wrote my solutions to challenge 74, 2 quite easy questions

dcw803:
sorry for missing last year's 100-submission challenge, once again I was busy last week (with actual Perl programming, gasp).  I'm not quite sure why "First" Non-repeating character in eg 1, pass 2 "ab" is b not a, seems to me that b is the "last" non-repeating char in ab.  but anyway, I get your output..


manwar:
Thanks @dcw803

PR #2143, opened 2020-08-25
Challenge 1

Scimon:
This does one thing differently, It shows the result in order of smallest number of coins to largest. I just think it looks nicer :)


manwar:
Thanks @Scimon

PR #2146, opened 2020-08-25
Week 75 soultions

drbaggy:
I've changed back to my main git account... just to confuse slightly...

The coin counting problem is really quite interesting...!


manwar:
Thanks @drbaggy

PR #2150, opened 2020-08-27
Branch for challenge 075

waltman:
This is my Perl and Python code for challenge 75. There's also a link to a blog post.


manwar:
Thanks @waltman

PR #2179, opened 2020-08-30
incorporated my solution of this week's tasks.. 2 mildly tricky questions

dcw803:
..especially the second one which made me think quite hard..  I suspect there's a more elegant algorithmic approach, but I went with what seemed reasonably obvious..


manwar:
Thanks @dcw803

PR #2210, opened 2020-09-05
pwc076 / task 2

xkr47:
Learned a lot new about Rust again, great task :)


manwar:
It gives me immense pleasure when the tasks is liked by the members. Thanks @xkr47

PR #2211, opened 2020-09-05
Pwc076 task 2 bugfix

xkr47:
I failed to traverse the whole grid in case width != height, which it was! Doh! Now it finds "hazard" also :)


manwar:
Thanks @xkr47

PR #2218, opened 2020-09-06
improved and explained

holli-holzer:
I added a small 4x4 grid example and a version of the script with verbose output for you Mohammad. 
It should be easy understand now what it does and how it works.

Just run `perl6 ch-2.explained.raku ch-2.words.small ch-2.board.small`


manwar:
I will check out on Monday morning after releasing the weekly challenge. I want to get to the bottom and wrap my head around it.
Thanks for your help. @holli-holzer

PR #2225, opened 2020-09-07
belatedly finished the second question and imported my solutions

dcw803:
after I fell asleep at the keyboard around 11pm:-) sigh!


manwar:
I have done this once during one Sunday night while release the weekly challenge. Thanks for sharing your solutions @dcw803

dcw803:
That fits us sad nerdy programmers to a tee - slaving over a hot keyboard until
we drift off to sleep:-) - and not even for work in this case.  Thanks for accepting
my sleepily delayed submission..

Best Wishes
Duncan

________________________________________

PR #2243, opened 2020-09-09
77-2: a (semi-)functional solution in XSLT

ash:


manwar:
Cool, BTW how do I run the script? Thanks @ash

ash:
It's said there in the XML example.

You need `xsltproc` or any other XSLT processor, and then: `xsltproc ch-2.xslt ch-2.xml `

PR #2246, opened 2020-09-10
Add challenge-077's ch-1 incomplete solution in Perl

andinus:
It's incomplete because it will not print all possible sums.

I had written this on the same day the challenge was posted & thought
I'll work on it later but it's Thursday & I don't think I'll work on
it this week so I'm pushing it in this state, maybe will complete it
next week.


manwar:
Thanks @andinus

PR #2253, opened 2020-09-11
77-2: update the solution to make less cell checks

ash:
ch-2a.cpp is a copy of ch-2.cpp but I added the counter that counts the number of checks that the program makes for testing the neighbours.

The second version, ch-2b.cpp, is the modified program that makes less checks if there are neighbouring cells with X.


manwar:
Thanks @ash

PR #2275, opened 2020-09-13
imported my solutions for this week's questions

dcw803:
I'm sure there's a more elegant solution to the first one.  Glad to see you clarified that, this time round, no repeats were alllowed.


manwar:
Thanks @dcw803

PR #2277, opened 2020-09-14
Corrections for challenge 076

therealjcm:
I realized I had searched for words 4 and longer in my submitted solution. Also I had repeated loop that could be simplified.


manwar:
Thanks @therealjcm

PR #2284, opened 2020-09-14
#78 solutions... kept nice and short!

drbaggy:
Nice fun little challenges - good as they show where Perl works really well..!


manwar:
Thanks @drbaggy

PR #2315, opened 2020-09-18
Solution for challenge 78 - 1

vinodk89:
Solution for challenge 78 - Task 1


manwar:
Thanks for your first contributions. @vinodk89

PR #2316, opened 2020-09-18
Solutions to PWC078 by Flavio Poletti (polettix)

polettix:
One of the two blog items will be published tomorrow morning... For the rest it should be OK :-)


manwar:
Thanks @polettix

PR #2325, opened 2020-09-20
submission for Challenge 078 and update 077

E7-87-83:
submission for the latest challenge: 2 Perl scripts


manwar:
Happy to see you are back on track. Thanks for sharing your work. @E7-87-83

PR #2331, opened 2020-09-20
Solution to week 78 by juliodcs

juliodcs:


manwar:
Welcome on board and thanks for your first contributions. @juliodcs

PR #2333, opened 2020-09-21
Challenge 78 by Jaldhar H. Vyas

jaldhar:
Late I know, but I want to get back on track after another forced hiatus.


manwar:
Thanks for sharing your solutions. @jaldhar

PR #2336, opened 2020-09-21
Solutions for challenge 79

davorg:


manwar:
Thanks for sharing your solutions. @davorg

PR #2345, opened 2020-09-22
Challenge 79

drbaggy:
Solutions to the two challenges - an O(ln(n)) solution for the first and a straightforward solution to the second - but added water levels to help visualise the solution.


manwar:
Thanks @drbaggy

PR #2347, opened 2020-09-22
[ch-079/jeongoon] change the name ch-1.raku ch-1.short.raku

jeongoon:
my ch-1.raku has limitation, I'll code tomorrow. :-)


manwar:
Thanks @jeongoon

PR #2359, opened 2020-09-23
pwc079.1

xkr47:
In Rust, works for any n = 0 .. 2¹²⁸−1
Naive implementation included for fun and for verifying the optimized version works :)


manwar:
Thanks @xkr47

xkr47:
Actually only works up to like 2¹²° because of overflow.. would need to do full mod-1000000007 math to handle the higher values.

PR #2376, opened 2020-09-25
My solution for Challenge 079 - Task #1.

rakulius:
My first proposal.


manwar:
Welcome on board and thanks for your first contribution. @rakulius

PR #2384, opened 2020-09-27
[ch-079/jeongoon] add ch-1.hs, just little bit more better ch-1.pl, ch-1.raku

jeongoon:
Thank you~~!!


manwar:
Thanks @jeongoon

PR #2399, opened 2020-09-28
solutions of PWC 080 from gugod in Perl, Raku, and Kotlin.

gugod:
I also wrote a blog about this:

https://gugod.org/2020/09/solutions-of-pwc-080/


holli-holzer:
great observation in #2. but for #1 check out the[ set operators
](https://docs.raku.org/language/operators#Junctive_OR_(any)_precedence)

manwar:
Thanks @gugod

PR #2406, opened 2020-09-29
prob 2 - Removed need for outer loop by reversing 2nd loop...

drbaggy:
Also added more complex test examples to check this - as the examples in the script can be "solved" by chance with incorrect assumptions...

Ch-1.pl also has a more complex test as part of a speed check against the other two solutions {using a hash and just doing multiple scans of an unsorted array}


manwar:
Thanks @drbaggy

PR #2417, opened 2020-09-30
Add solutions for week 80

tedleahy:
This is the second week I've done this challenge, and I just want to say thanks a lot, it's great! I'm fairly new to perl (been working with it for just under 2 months) so these are a lot of fun and really useful for improving my new perl skills.

Let me know if you notice any noob mistakes, I'm still not 100% sure about Perl best practices :)

-Ted


manwar:
You can learn a lot about best practices by just reading others solutions. I would suggest checkout Perl reviews every week. Also take a look at recaps. There are plenty to keep you busy all week. I hardly find time but I keep them bookmarked for leisure reading. @tedleahy

PR #2426, opened 2020-10-01
Add polettix's solution to PWC080

polettix:
Blog entry for count candies is still not out... but will be soon ;-)


manwar:
Thanks @polettix

PR #2466, opened 2020-10-06
Fix bug on challenge-080 ch-1, compact challenge-081 ch-2

andinus:
This fixes the bug in challenge-080 ch-1 where repeated elements would
break the code & removes a useless line in challenge-081 ch-2.


manwar:
Thanks @andinus

PR #2480, opened 2020-10-09
WordCount

frankivo:
Scala solution for the Word Count.

Tried to make it a oneliner 🥇


frankivo:
Code can be executed by using ``scala WordCount.scala``.
I could also provide an sbt-environment if you would prefer that.

manwar:
Welcome on board and thanks for your first contributions. @frankivo

PR #2482, opened 2020-10-09
Ch81/Task 2: optimize and add comments

tylerw:
We just need a count of the words, which can be achieved by summing (via
x/reduce +) 1 for every instance we find.

Add comments to clarify what the algorithm is doing at each step.


manwar:
Thanks @tylerw

PR #2503, opened 2020-10-12
sgreen solution for challenge 082

simongreen-net:


manwar:
Thanks @simongreen-net

PR #2511, opened 2020-10-13
Add entries for 082/polettix

polettix:
Blog entry for task #2 is on its way for being published. Thanks for all the energy you put!


manwar:
Thanks @polettix

PR #2522, opened 2020-10-14
[gugod] an imprement of 082.1

gugod:
Hi,

I updated my solution and also wrote about them here

- https://gugod.org/2020/10/pwc-082-raku-en/
- https://gugod.org/2020/10/common-factors-and-gcd-en/


manwar:
Thanks @gugod

PR #2524, opened 2020-10-14
Improve the performance of week 082/part 1.

Abigail:
Improved the running time of the algorithm from O (sqrt (gcd ($M, $N)))
to O (sqrt (p) + |f| log |f|) where p is the largest prime factor
of gcd ($M, $N), and |f| is total number of factors.
This is not an improvement if the gcd is a prime, but a dramatic
improvement if the gcd contains lots of prime factors.


manwar:
Thanks @Abigail

PR #2525, opened 2020-10-14
my aproach to challenge 082

jluis:
This week  I cannot thick of a fancy solution so I push this as WIP solution


manwar:
Thanks @jluis

PR #2548, opened 2020-10-18
Adjust solution to examples :)

polettix:
Ehr... sorry for all the noise, but I solved a different problem in challenge 2, and I had to change it. I also added another blog post about this... I hope it can be useful.

Sorry again for the trouble and thanks for all that you do!


manwar:
Thanks @polettix

PR #2552, opened 2020-10-18
[ch-082/jeongoon] ch-2.pl use two solution (quick check, show all cases), ch-2.go changed to use interface.

jeongoon:


manwar:
Thanks @jeongoon

PR #2561, opened 2020-10-18
imported my answers to both questions; the 2nd was surprisingly easy …

dcw803:
…although I found it hard to think about before I started, because I didn't find the definition of "interleave" clear enough


manwar:
Thanks @dcw803

PR #2563, opened 2020-10-19
[ch-082/jeongoon] I think this is correct one {ch-2.go, Ch2PlanA.pm}

jeongoon:
Sorry, I know it is late. but this version is more accurate.


manwar:
Thanks @jeongoon

PR #2573, opened 2020-10-19
This week solutions 

jluis:
I would like to put the fist as a one-liner but I didn't find a reference of how it can be better integrated on the repo:


manwar:
Thanks @jluis

PR #2598, opened 2020-10-23
[ch-083/jeongoon] add more versions of Task #2 for Perl, Haskell, and another blog for Task #2

jeongoon:


manwar:
Thanks @jeongoon

PR #2625, opened 2020-10-26
imported my solutions, slightly late, only started at 11:40pm:-)

dcw803:
I didn't have much time to think about the second one, so my solution is not very elegant; solid, does the job, no inspiration, no cleverness, probably not very efficient


manwar:
Thanks @dcw803

PR #2635, opened 2020-10-26
Fix incorrect implementation of combinations for challenge-083-2-perl

mienaikage:
I might come back and refactor this, but in the meantime this fixes my previously incorrect handling of possible combinations.


manwar:
Thanks @mienaikage

PR #2636, opened 2020-10-26
challenge solutions

drbaggy:
This weeks were quite easy - and so the coding is a bit naive - and I can't see any real "optimizations" in this code..


manwar:
Thanks @drbaggy

PR #2640, opened 2020-10-26
Challenge 84 Solutions (Raku)

andemark:


holli-holzer:
I find #2 a bit from behind trough the chest into the eye, but #1 is really clever. hat off.

andemark:
I made a slight change to both. Better or worse?
BTW, your "[X] map { +$_, -$_ }, @A" from last week was awesome. I gotta
remember that one.

manwar:
Thanks @andemark

PR #2660, opened 2020-10-30
initial commit on challenges 79-82

stuart-little:


manwar:
Thanks @stuart-little

PR #2673, opened 2020-11-01
Challenge 084

ccntrq:
My solutions for the current challenge. 

There are also minor tweaks to last weeks solution so that i could easier reuse the `combinations` sub.


manwar:
Thanks @ccntrq

PR #2684, opened 2020-11-02
After the first commit with naive solutions - added some optimizations to allow this to work with larger numbers...

drbaggy:


manwar:
Thanks @drbaggy

PR #2685, opened 2020-11-02
Challenge 85

Scimon:
There's probably a super maths way of doing the second challenge.

I am so tired I couldn't even read it correctly though.

I hope you're OK. :)


manwar:
I am doing fine now, thanks. @Scimon

PR #2698, opened 2020-11-03
Add polettix's solution to challenge 085

polettix:
Hi, thanks for all your hard work.

Blog entries will be published in due time :smile:


manwar:
Thank you for your support and contributions. @polettix

PR #2703, opened 2020-11-04
Add solution to 085 Triplet Sum & Power of Two Integers (E. Choroba)

choroba:
The Triplet Sum solution is the fastest I could find.


manwar:
Thanks @choroba

PR #2729, opened 2020-11-08
committed my solutions to this week's questions

dcw803:
I couldn't think of a very interesting approach, I'm afraid


manwar:
Thanks @dcw803

PR #2831, opened 2020-11-23
First pass at ch 88

drbaggy:
Nice challenges this week....

I like the fact that you don't need to worry about indicies in the 2nd example - just knowing when you can shift or pop elements!


manwar:
Thanks @drbaggy

PR #2855, opened 2020-11-27
Add polettix's solution to PWC088

polettix:
Hi, thanks for all the hard work.

This is my solution for this week's challenge. Blog posts are already in place.

Thanks!


manwar:
Thanks @polettix

PR #2866, opened 2020-11-29
[ch-088/jeongoon] Perl, Raku, Haskell, Go, Common-lisp, Elm solution added.

jeongoon:
Thank you~!
See you next year!!


manwar:
Thanks @jeongoon

PR #2878, opened 2020-11-29
imported my solutions; liked both tasks

dcw803:
especially the spiral path one which made me think…


manwar:
Glad you liked the tasks. Thanks for sharing solutions. @dcw803

PR #2891, opened 2020-11-30
- Solution for Weekly Challenge 089 by Xliff

Xliff:
My first solution. This was fun! Thank you for organizing these!


manwar:
Welcome on board and thanks for your first contributions. @Xliff

PR #2914, opened 2020-12-04
Change "up" link in all README.org

andinus:
These are not solutions but just changing the `HTML_LINK_UP' property
of all README.org files. I changed the website layout.


manwar:
Thanks @andinus

PR #2929, opened 2020-12-06
Add solutions by E. Choroba to 089: GCD Sum & Magical Matrix

choroba:
GCD Sum shows both the naive slow algorithm and the fast one based on
Euler's totient function (studied at
https://www.geeksforgeeks.org/summation-gcd-pairs-n/).


manwar:
Your contribution is priceless. Thanks @choroba

PR #2952, opened 2020-12-08
Java Solution for Ethiopian Multiplication

ziameraj16:
This is a Java solution to Ethiopian Multiplication


manwar:
Welcome on board and thanks for your first contributions. @ziameraj16

PR #2974, opened 2020-12-13
1st commit on 024

stuart-little:


manwar:
Thanks @stuart-little

manwar:
ch-1.p6 is empty? @stuart-little

stuart-little:
> ch-1.p6 is empty? @stuart-little

Sure: [the problem](https://perlweeklychallenge.org/blog/perl-weekly-challenge-024/) asks for the shortest script that doesn't return an error :)

PR #2984, opened 2020-12-14
Solutios week 91

drbaggy:
This weeks were really nice - a little bit of thinking required - the first one is nice as you can use a one-liner "regex" regex....

- **r** -> return replaced string (rather than updating it)
- **e** -> execute the code in the replace string (rather than a direct replacement)
- **g** -> execute multiple times (ok most people now this one!)
- **e** -> superfluous - didn't need to put it in - but it makes the "trick" look better!
- **x** -> add whitespace and comment the regular expression to make it more readable


manwar:
Thanks @drbaggy

PR #3011, opened 2020-12-19
091 gforth solution

pauloscustodio:


manwar:
Thanks @pauloscustodio

PR #3014, opened 2020-12-20
Challenge 090 solution in gforth

pauloscustodio:


manwar:
Thanks @pauloscustodio

PR #3026, opened 2020-12-20
imported my attempts at this week's puzzles

dcw803:
I thought the jump-game would be more challenging than it was; with negative values, and the ability to start at any position (not just pos 0). it might be quite cool..


manwar:
Thanks @dcw803

PR #3032, opened 2020-12-21
092-1 in Raku by ash

ash:


stuart-little:
I don't think this does then intended task @ash, if I interpret [isomorphism](https://www.educative.io/edpresso/how-to-check-if-two-strings-are-isomorphic) as in that link:

All they ask is that the chars in one string be mappable one-to-one to the chars of the other. So for instance you should return 1 for `ACAB` and `XCXY` (the link actually gives that example). 

Your script returns 0 for those though, because you're asking for a more stringent condition: you want the differences between codepoints to be equal, which is much more than isomorphism.

ash:
Yeah, you may be right.

stuart-little:
Yeah.. It's funny how Mohammad's [examples](https://perlweeklychallenge.org/blog/perl-weekly-challenge-092/#TASK1) still fit both notions of isomorphism though, as far as I can tell by sheer happenstance: 'a', 'b' and 'c' are consecutive, as are 'x', 'y' and 'z'.

I had to actually go to that [other link](https://www.educative.io/edpresso/how-to-check-if-two-strings-are-isomorphic) to confirm 'isomorphism' meant what I thought it did.

manwar:
Thanks @ash

PR #3035, opened 2020-12-21
initial

andemark:


stuart-little:
@andemark I believe your `ch-1` script doesn't check whether the two strings have the same length, and the zip simply cuts off the tail of the longest.

So `abca` and `xyz` register as isomorphic (I've tried) even though they're not..

andemark:
> @andemark I believe your `ch-1` script doesn't check whether the two strings have the same length, and the zip simply cuts off the tail of the longest.
> 
> So `abca` and `xyz` register as isomorphic (I've tried) even though they're not..

Thanks for pointing that out, Stuart. I think it's all working now :)

stuart-little:
Sure looks good to me :).

manwar:
Thanks @andemark

PR #3036, opened 2020-12-21
Add solutions to 092: Isomorphic Strings & Insert Interval

choroba:
Insert Interval can be solved by using the code form 050/1 Merge
Intervals which I did.


manwar:
Thanks @choroba

PR #3038, opened 2020-12-22
challenge 92, 1 & 2 - in raku

pkmnx:


stuart-little:
I do not believe your challenge-1 script works as intended @pkmnx : what is required is that one be able to transliterate the 1st string into the 2nd upon applying a one-to-one map on letters.

So for instance 'aac' and 'xyx' should *not* count as isomorphic, because the 1st has coinciding letters 1 and 2 while the 2nd doesn't.

Your script returns 1 for these two strings though, because it only takes into account the various letter counts. So all strings of 3 letters where two letters coincide count as isomorphic for your script, regardless of *where* the two coinciding letters sit.

stuart-little:
Incidentally, my own initial commit broke on the same example (of `aac` and `xyx`), so I've had to fix mine too :).

manwar:
Thanks @pkmnx

PR #3040, opened 2020-12-22
Completely different solution to ch-1.. which I think is incredibly neat

drbaggy:
I've left my original solution around - I've also written a script that goes through a dictionary and finds and displays all isomorphic words....


drbaggy:
Just added additional copy of the function in ch-1 without comments as I think it illustrates the "simplicity" of the method...

manwar:
Thanks @drbaggy

PR #3060, opened 2020-12-23
Remove need to call sqrt()

pauloscustodio:


manwar:
Thanks @pauloscustodio

PR #3074, opened 2020-12-27
1st commit on 027

stuart-little:
Last one for `raku`! I guess I'd better find another language to redo all of the challenges in :D.

Thank you for the patience and encouragement @manwar !


manwar:
Thanks @stuart-little

PR #3094, opened 2020-12-27

dcw803:
I ran out of time to do the intervals one in an elegant way, I know there is one but couldn't quite see it, so instead I did it the totally dumb way: build a number line: an array x[] of booleans, setting each x[i] to true when i is part of an interval, either an existing interval or the new one, then walk the @x array to produce the minimal number of intervals.  of course the @x array automatically does the merging for us.  this, btw, merges even more intervals than your examples do.


manwar:
Thanks @dcw803

PR #3108, opened 2020-12-29
Improve gustavo-chaves's challenge 093

gnustavo:
The max-points function now uses Math::BigRat objects in order to avoid
rounding errors as much as possible.


manwar:
Thanks @gnustavo

PR #3111, opened 2020-12-29
Add Perl solution to challenge 005

pauloscustodio:
and remove one-letter solutions from challenge 004 task 2


manwar:
Thanks @pauloscustodio

PR #3112, opened 2020-12-29
Add Perl solution to challenge 006

pauloscustodio:
and fix way of getting PI with n digits in challenge 004.1


manwar:
Thanks @pauloscustodio

PR #3140, opened 2021-01-03
committed my solutions to this week's tasks..

dcw803:
first task interesting, I felt I wasn't seeing an elegant way to do it, but easy enough to do it the brute force way; the second task looks like a retread of challenge 56's task 2, so I reused most of the binary tree representation and input parsing from last time..  nice..


manwar:
Thanks @dcw803

PR #3148, opened 2021-01-04
Update ch-1.pl of Challenge 093

E7-87-83:
The time complexity of the case should be O(n log n), not as ideal as O(n)


manwar:
Thanks @E7-87-83

PR #3183, opened 2021-01-08
Challenges

wlmb:
I made a small change the the program for challenge 94, task 2, ch-2.pl, but I tried an alternate solution which I named ch-2a.pl. I don't know if there is some standard naming scheme when more than one solution to a task is submitted. 
Regards,
Luis


manwar:
Thanks @wlmb

PR #3207, opened 2021-01-10
Add Python solution to challenge 093

pauloscustodio:


manwar:
Thanks @pauloscustodio

pauloscustodio:
Thank you Mohammad for organizing this contest. It is fun to try to solve
the problems first in Perl, and then in tougher languages like Forth, which
is very low level. It is also a good opportunity to get some practice on
languages I have rarely used like Python.

Regards,
Paulo Custódio

PR #3210, opened 2021-01-10
Mimosinnet raku solution

mimosinnet:
Thanks for the challenge! I have really struggled with the second one and I have had to use Luca Ferrary to be able to solve it. Superb Raku exercise! Thanks!


manwar:
Thanks @mimosinnet

PR #3213, opened 2021-01-10
imported my solutions to challenge 94

dcw803:
loved the anagram groups task as good old Bell Labs alumni Jon Bentley tackled precisely this problem in the wonderful book "Programming Pearls".  Core concept: signatures: the sorted bag of letters contained in a word.  All words that contain the same letters necessarily have the same signature.


manwar:
Thanks @dcw803

PR #3242, opened 2021-01-13
Add ch-1 in nim

mfoda:
Solution file is named `ch1.nim` as nim doesn't allow hyphens in module/file names.


manwar:
Welcome on board and thanks for your first contribution. @mfoda

PR #3305, opened 2021-01-18
account for invalid dates in 019 ch-1

stuart-little:


manwar:
Thanks @stuart-little

PR #3319, opened 2021-01-18
updated prolog code for part 1

adamcrussell:


adamcrussell:
the original code works fine, this update just makes it work in "two directions". now if you call palindrome/1 with an uninstantiated variable it will generate a palindromic number. As originally written it just validates whether a given number is palindromic or not.

manwar:
Thanks as always for support and encouragement. @adamcrussell

PR #3332, opened 2021-01-21
Solutions for challenge 096

mimosinnet:
I am thrilled to have been able to solve the second exercise with Raku OOP. Thanks a lot for the challenges! It is an excellent way to learn and practice raku.


manwar:
Thanks @mimosinnet

PR #3346, opened 2021-01-22
Add Perl solution to Challenge 022

pauloscustodio:


manwar:
Thanks @pauloscustodio

PR #3349, opened 2021-01-22
Replace Math::Prime::Util by explicit prime number generation

pauloscustodio:


manwar:
Thanks @pauloscustodio

PR #3367, opened 2021-01-24

dcw803:
task 1: trivial, task 2: only did the naive recursive form cos I got bogged down in trying to think up how to do it from scratch.  The iterative versions looked rather clever.


manwar:
Thanks @dcw803

PR #3386, opened 2021-01-26
Add Lua solution to 096

pauloscustodio:


manwar:
Thanks @pauloscustodio

PR #3412, opened 2021-01-30
Rename Ch1 to Perl5

LubosKolouch:


manwar:
The parser accepts both "perl" and "perl5" folder name. Thanks @LubosKolouch

PR #3425, opened 2021-01-31
correction and improvement

E7-87-83:
093 task 1: correction
097 task 2: removal of unnecessary variables


manwar:
Thanks @E7-87-83

PR #3444, opened 2021-02-01
1st commit on 062_haskell

stuart-little:
Last one in Haskell!


manwar:
Thanks @stuart-little

PR #3449, opened 2021-02-02
Fix wrong links in comments for all programs

wlmb:
I just noticed I haven't been using double slashes in the comments that link to the corresponding blog pages. Fixed that.
Regards


manwar:
Thanks @wlmb

PR #3483, opened 2021-02-08
Pwc99

fluca1978:
My raku solutions.
The second task is not fully working, but I cannot find a better solution right now.


manwar:
Thanks @fluca1978

PR #3486, opened 2021-02-08
Add solution to 099 Pattern Match & Unique Subsequence by E. Choroba

choroba:
Two different solutions are given for the Pattern Match: One
transforms the pattern into a regex, while the other really implements
the searching for the pattern character by character.

Similarly, there are two subroutines in the Unique Subsequence: One
just returns the final count of solutions, the other one enumarates
all the valid positions in the string.


manwar:
Thanks @choroba

PR #3488, opened 2021-02-08
Java solution for the pattern match

ziameraj16:


xkr47:
Hi, just noticed that Task #1 asked for a wildcard pattern match, not a regular expression pattern match that you have..

ziameraj16:
Good point, Will `?` be considered a wildcard as well?

xkr47:
As defined in the task, `*` and `?` are the only wildcards.

manwar:
Thanks @ziameraj16

PR #3493, opened 2021-02-09
9 and 90 challenges...

jacoby:


challenge-099/dave-jacoby/perl/ch-1.pl
```diff
@@ -0,0 +1,29 @@
...
+    my $p = pattern_match( $n->@* );
+    say $n->[0];
+    say $n->[1];
+    say $p;
+}
+
+sub pattern_match ( $S, $P ) {
+    my $pattern = $P;
+    $pattern =~ s/\*/.*/g;
+    $pattern =~ s/\?/.?/g;
```
xkr47, line 25: Hi, just happened to notice a bug here :)
```suggestion
    $pattern =~ s/\?/./g;
```
How about characters "." and "\" that are special in regexp but not in these patterns?

jacoby: Thanks for that.

manwar:
Thanks @jacoby

PR #3506, opened 2021-02-12
tidied up and docs and blog links

drbaggy:
Add blog links, add some comments - and include timings to show speeds of each fn.


manwar:
Thanks @drbaggy

PR #3507, opened 2021-02-13
some better golfing - over 30% shorter solution for ch-2.pl

drbaggy:
Re-golfed challenge 2 from week 97.

The body of the function for ch-2.pl golfed so the code inside the function has been reduced to 145 characters from around 200


manwar:
Thanks @drbaggy

PR #3517, opened 2021-02-14
Add Awk solution to challenge 004

pauloscustodio:
Fix Awk syntax on previouws submissions - semicolon is needed at the end of statements


manwar:
Thanks @pauloscustodio

PR #3521, opened 2021-02-14

dcw803:
I loved both questions (and tried implementing task 1 in two different ways, a simple way at first, then later a let's do it from scratch way..)


manwar:
Thanks @dcw803

PR #3530, opened 2021-02-15
My solution to challenge 99/task#1

x1mandi:
Hi Mohammad,

I apologize for the delay, this is my solution for last weeks challenge task#1.

I hope You will still accept it.

Kind regards,
Csaba


manwar:
It is never too late. Thanks @x1mandi

PR #3535, opened 2021-02-15
clean up ch-1 corner cases

stuart-little:


manwar:
Thanks @stuart-little

PR #3542, opened 2021-02-16
Save a few bytes by trimming the subroutine name, and using pop rather than shift! and removing extraneous ;

drbaggy:


manwar:
Thanks @drbaggy

PR #3555, opened 2021-02-17
Added additional functions to challenge 2

drbaggy:
My rewriting the inner for loop as a map, can rewrite challenge 2 function as a 1.5 liner {1 to do the donkey work} + one to return the correct value...

Generate a simple function to store the paths through the triangle while working backwards and then displaying them with square brackets around the numbers chosen to generate the minimal value...


manwar:
Thanks @drbaggy

PR #3569, opened 2021-02-19
Added solution for ch100

Doomtrain14:
Hello :D
Added perl solution for ch100


manwar:
Good to see you back. Glad you manage to submit PR finally. Thanks @Doomtrain14

PR #3591, opened 2021-02-21
Add Ada solution to challenge 096

pauloscustodio:
Simplify all the other solutions: output only the number of steps, as requested, and not the list of steps.


manwar:
Thanks @pauloscustodio

PR #3597, opened 2021-02-21
imported my solutions to ch#100, and also did a lot more work on ch#99s..

dcw803:
pattern matching - ch#99's README is basically now a blog (I've never submitted a blog before, because I don't have a blog set up).  I might convert this README into an article sometime when I have time.


manwar:
Thanks @dcw803

PR #3645, opened 2021-03-01
Solution to challenge 2 - currently optimizing challenge 1

drbaggy:


manwar:
Loved your dedication to keep improving. Thanks @drbaggy

PR #3658, opened 2021-03-04
First one is nasty this time {well at least trying to find an efficient solution}

drbaggy:
Done some optimization - still in the order-of-magnitude of minutes for the more complex ones...

Applied a lot of filtering based on possible values for some of the digits, and the fact that the digit sum is restricted to certain values

Lots of screaming!...!


manwar:
Thanks @drbaggy

PR #3672, opened 2021-03-06
Solution for challenge 102

mimosinnet:
Thanks for the challenge. Learned a lot with challenge 1, and I had to look at other solutions to be able to solve it. Still, I am unsure this is the right approach. Good to practice recursion with challenge 2. Cheers!


manwar:
Thanks @mimosinnet

PR #3684, opened 2021-03-07
imported my solution's to this week's tasks

dcw803:
SEE ALSO OptimizingTask1 for an account of a series of profiling-driven optimizations that I made to speed up finding rare numbers.  Overall, my fastest version ran approx 18 times faster than the original.


manwar:
Thanks @dcw803

PR #3712, opened 2021-03-13
Add elm solution

lancew:
Cheating a little @manwar as this is just a module and not console executable.


manwar:
Thanks @lancew

PR #3720, opened 2021-03-14
Imported my solutions to challenge 103

dcw803:
(btw, submission date on main web page was wrong, said march 8th)

Two relatively easy tasks this week, quite relaxing


manwar:
Thanks @dcw803

PR #3741, opened 2021-03-18
Add solutions to 104: FUSC Sequence & NIM Game by E. Choroba

choroba:
Also, include a graphviz source of a diagram showing all the possible
states of the NIM Game.


manwar:
Thanks @choroba

PR #3756, opened 2021-03-21
imported my solutions to this week's tasks.. two nice questions

dcw803:
but please note that TASK 2 WAS NOT NIM.  It's the "simple take away game".  Nim is played on multiple piles, with a single move being to take away ANY NUMBER of tokens from a single pile, and the winning solution involves reducing the xor-sum of the number of tokens on all piles to zero after your move.


manwar:
Thanks @dcw803

PR #3758, opened 2021-03-22
fix a serious bug! ... after deadline

E7-87-83:
no wonder why the seq terms were such small from my code


E7-87-83:
I will do better onward

manwar:
Thanks @E7-87-83

PR #3760, opened 2021-03-22
added in first pass

drbaggy:
1.  not using ** or log, but a "modified" binary search...
2.  with starting vowels & consonant and b/f/m rules applied


manwar:
Thanks @drbaggy

PR #3767, opened 2021-03-22
(update to challenge 102): fixed a bug in ch-1i.pl and added a mention of OptimizingTask1 to ch-1a.pl

dcw803:
:-)


manwar:
Thanks @dcw803

PR #3791, opened 2021-03-28
imported this week's solutions

dcw803:
not much to get excited about this weeks, nthroot was simple but dull, and the children's rhyme has no real clear rules, so doesn't deserve a deep solution


manwar:
Thanks @dcw803

PR #3809, opened 2021-03-31
Add polettix's solution to challenge-106

polettix:
Hi! Please find my takes on this week's challenges.

Thank you very much for all the hard work :grin:


manwar:
Thanks @polettix

PR #3838, opened 2021-04-05
Add solution to 107: List Methods by E. Choroba

choroba:
Reuse 043 for the solution of Self-descriptive Numbers.


manwar:
Thanks @choroba

PR #3843, opened 2021-04-06
More notes on both ch-1.pl & ch-2.pl 

drbaggy:
Using some "perlisms" to reduce the code for ch-1.pl to a "single" statement - and then some golfing to run it as compactly as possible in a command line 1-liner!

Demonstrating the difference between use/require and which methods are not imported with require. require gives the same list as the embedding the class in the perl script, whereas use gives the extra fn import!


manwar:
Thanks @drbaggy

PR #3848, opened 2021-04-08
Add polettix's solution to challenge-107

polettix:
Hi! Thanks for keeping this alive since so much time!


manwar:
Thanks @polettix

PR #3855, opened 2021-04-10
Solutions for challenge 107

mimosinnet:
Thanks very much for the challenges. Task #2 on methods and classes has been quite formative for those who are not so used to classes and methods. Great!


manwar:
Thanks @mimosinnet

PR #3890, opened 2021-04-13
challenge 002 finished with perl

lakpatashi:


manwar:
Thanks @lakpatashi

manwar:
I realised it is same as challenge 001. 
Did you ,miss anything?

PR #3894, opened 2021-04-14
Finished challenge-004 with perl

lakpatashi:
first commit


manwar:
Incredible speed. Thanks @lakpatashi

PR #3898, opened 2021-04-15
I realized the code wasn't as clean as I wanted it to be.

PerlBoy1967:
Bit of a rush job between some real (paid) work.
- Added commandline argument
- Make use of array to store the triangle
- Do all index math using $_ or magical indexes (-1,-2)


manwar:
Thanks @PerlBoy1967

PR #3972, opened 2021-04-27
1+2 - plus some text fixes from last week...

drbaggy:
Also some notes about different approaches to 2 - and why naive isn't always best!


manwar:
Thanks @drbaggy

PR #3976, opened 2021-04-28
Added a minimalistic solution for Task 2

PerlBoy1967:
Based on solutions provided by other teammembers their '[split/,/]' usage.


manwar:
Thanks @PerlBoy1967

PR #3992, opened 2021-05-02
 remove unnecessary line on code of wk 109; Submission for week 110 

E7-87-83:
Perl scripts & Java for both tasks


manwar:
Thanks @E7-87-83

PR #4022, opened 2021-05-05
Move from 'unpack' back to split('').

PerlBoy1967:
After some benchmarking I came to the conclusion split is faster.


manwar:
Thanks @PerlBoy1967

PR #4041, opened 2021-05-09
Solutions for challenge 111

mimosinnet:
As David Shepherd, I wouldn't keep both hands on the keyboard while challenge 111 is on the board. Thanks for so much fun! :)


manwar:
Thanks @mimosinnet

PR #4046, opened 2021-05-10
dms061 - Added words file used by ch-2

dms061:
Forgot to add the words file my ch-2.pl script uses to answer question 2.


manwar:
Thanks @dms061

PR #4071, opened 2021-05-12
Solution 1 to challenge 112

kurkale6ka:


manwar:
Welcome on board and thanks for your first contributions. @kurkale6ka

PR #4096, opened 2021-05-17
First pass at challenge 113

drbaggy:


challenge-113/james-smith/perl/ch-1.pl
```diff
@@ -0,0 +1,23 @@
...
+
+use strict;
+
+use warnings;
+use feature qw(say);
+use Test::More;
+use Benchmark qw(timethis);
+
+my @ex = ( [25,8,0], [25,7,0], [24,7,1], [24,0,0], [10,0,1], [28,8,1], [26,8,1], [16,8,0], [441,9,1], [431,9,0] );
+
```
simongreen-net, line 11: 16 can be made from 8 + 8, 431 and 9 can be made from 395 + 9 + 9 + 9 + 9.

drbaggy: Actually need to look into that 16 - the question suggests that it is different numbers... Although will need to look into the last one as it can be written as 295 + 9 + 19 + 29 + 79 ....

drbaggy: Found the solution - which was to check for numbers larger than (100) for 0 & for numbers larger than 10$d for other digits as there is a systematic way to make these up with two numbers - 10{x} + {y}0 or similar for other numbers...

manwar:
Thanks @drbaggy

PR #4114, opened 2021-05-20
Revise Ben Davies' solution to part 2 of challenge 113

Kaiepi:
Nits as far as the challenge is concerned.


manwar:
Thanks @Kaiepi

PR #4124, opened 2021-05-23
Add polettix's solutions for Raku

polettix:
Hi! I decided to give Raku a try... and here's my almost-literal translation of the Perl code into it. It's a start!

Thank you for your hard work, Flavio.


manwar:
Thanks @polettix

PR #4156, opened 2021-05-28
Solutions for challenge 114

mimosinnet:
Done! Fascinated by the possibility of writing code like this in Raku. 
`($number^..^∞).grep({ $_ eq $_.flip }).first;`
Thaks!


manwar:
Thanks @mimosinnet

PR #4168, opened 2021-05-30
Add solutions and post

pflanze:
Only one solution due to lack of time.


manwar:
I am glad you made it. Please don't break the continuity. Thanks @pflanze

PR #4174, opened 2021-05-31
sgreen solution to challenge 115

simongreen-net:
Will write blog tomorrow.


simongreen-net:
Actually I've just thought of an edge case that means my code isn't correct.

PR #4209, opened 2021-06-06

dcw803:
two quite easy tasks, I thought the first one was going to be harder than it actually was, before realising that "all N words must be used in the chain", and that because it's circular, you can start with any word.


manwar:
Thanks @dcw803

PR #4212, opened 2021-06-06
Add python solution to challenge 004

pauloscustodio:


manwar:
Welcome back and thanks @pauloscustodio

PR #4234, opened 2021-06-10
Optimizations

drbaggy:
Added some code so more efficient in large number cases!


manwar:
Nice work, thanks @drbaggy

PR #4261, opened 2021-06-14
This weeks challenge

Scimon:
I might tweak challenge two to make it do a level 10 triangle in a reasonable time.


manwar:
Thanks @Scimon

PR #4294, opened 2021-06-19
Add C and C++ solution to challenge 005

pauloscustodio:


manwar:
Thanks @pauloscustodio

PR #4297, opened 2021-06-20
Challenge 117 (Part 1) - Raku

aaronreidsmith:
I only had time for part 1 this week; maybe I will come back and do part 2, but figured I would submit what I have! 🎉


manwar:
Good plan, thanks @aaronreidsmith

PR #4301, opened 2021-06-20
Add solutions to 115: Missing Row & Find Possible Paths

choroba:
In the POD of the second task, I explain why the task origianlly
requested the number of paths instead of their list.


manwar:
Thanks @choroba

PR #4318, opened 2021-06-22
Add solutions to 118: Binary Palindrome & Adventure of Knight

choroba:
The second task is solved using A* with a simple heuristic: the number
of treasures to pick up. Try commenting out the line that adds it to the
score to see its effect.


manwar:
Thanks @choroba

PR #4340, opened 2021-06-25
the codes are not in their best shapes

E7-87-83:
but maybe some people out there are eager and curious to the problem creator's approach


manwar:
Thanks @E7-87-83

PR #4345, opened 2021-06-26
Add a script to be run on CLI, repeats 99999 times 

lancew:
A relatively short path is found with 99999 repetitions, more required to get to my best-shortest route so far of 11 moves: (7,0) (6,2) (5,4) (4,2) (2,1) (0,2) (1,0) (2,2) (0,1) (1,3) (3,2) (1,1).

I might do a parallel processing experiment as this is single threaded, and I don't like that I have lost the A-H column names, as it goes against my original intent to make this express the problem without optimisations. :)

### DIFF
--- a/challenge-118/lance-wicks/perl/ch-2.pl
+++ b/challenge-118/lance-wicks/perl/ch-2.pl
@@ -1,0 +1,29 @@
+use strict;
+use warnings;
+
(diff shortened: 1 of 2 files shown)


manwar:
Great attempt. Thanks @lancew

PR #4355, opened 2021-06-28
imported challenge 118 solutions (hour and a half late)

dcw803:
second task (knight's moves) was surprisingly hard.. sorry I missed last week's, was away on holiday


manwar:
Thanks @dcw803

PR #4381, opened 2021-06-29
Update ch-1.pl

PerlBoy1967:
It appears 'slide()' changed over time??? 

The idea is still the same but hopefully this version is okay across versions of 'List::MoreUtils'.


manwar:
Thanks @PerlBoy1967

PR #4389, opened 2021-07-01
Challenge 119 

frankivo:
The second took me a few attempts before getting the functional programming way. Still lots of String conversions :(


manwar:
Thanks @frankivo

PR #4401, opened 2021-07-03
Simple (unsafe) solution to task 1

lancew:
I did a live stream too. :)
https://www.twitch.tv/videos/1075491044


manwar:
Thanks for live coding. @lancew

PR #4402, opened 2021-07-03
Quick GoLang solution to task 1

lancew:


manwar:
Wow, bonus for the week, thanks @lancew

PR #4404, opened 2021-07-03
Task 2 in Perl

lancew:
I thought that would be harder. Pretty simplistic approach but seems to work. :)


lancew:
Time for some blogging. ;-)

manwar:
Completed the weekly challenge, technically. Thanks @lancew

PR #4420, opened 2021-07-04
imported my solutions to challenge-119's two tasks..

dcw803:
.. both thankfully easy after last week's Knight's adventure question..


manwar:
Thanks @dcw803

PR #4507, opened 2021-07-13
Task 2 via Brute Force

E7-87-83:
Submit a brute-force solution; 
don't understand how to apply A* search, BFS, etc. (all advanced algorithms in wk 118) in this task;
maybe someone would like to have some handy test data for $N > 10 .


manwar:
Thanks @E7-87-83

PR #4516, opened 2021-07-14
Task 2 - Fun using brute force solution

PerlBoy1967:
Using libraries:
* MCE;
* Algorithm::Combinatorics qw(permutations);
* List::MoreUtils qw(slide);
* List::Util qw(sum min);


manwar:
Thanks @PerlBoy1967

PR #4554, opened 2021-07-18
sgreen solution to challenge 121, task 1

simongreen-net:
Only did one task this week.


manwar:
Thanks @simongreen-net

PR #4555, opened 2021-07-19
Challenge 121 part 1 by Jaldhar H. Vyas.

jaldhar:
Unfortunately I was really low on time so this is only the first challenge.  I'll try to do the other one tomorrow.


manwar:
Thanks @jaldhar

PR #4580, opened 2021-07-22
Challenge 122: Remove unnecessary method calls

andinus:
``.comb.grep(* !== "").join'' can be replaced with ``.join''.

And we don't have to convert it back to Int since we're just printing
it.


manwar:
Thanks @andinus

PR #4619, opened 2021-07-27
Task 2  - Check diffently distance frequency count

PerlBoy1967:
And add equilateral triangle check :: Thanks to James Curtis-Smith


manwar:
Thanks @PerlBoy1967

PR #4627, opened 2021-07-28
Add assumption that we are testing for points in a square not necessarily in the right order...

drbaggy:


manwar:
Thanks @drbaggy

PR #4682, opened 2021-08-08
imported solutions to this week's challenge tasks

dcw803:
I enjoyed the second task; I'm not at all that my way is "the best way", or even a "reasonably good way"; but it's a simple way to do:-)


manwar:
Thanks @dcw803

PR #4717, opened 2021-08-15
imported my solutions to this week's challenge - both quite tricky

dcw803:
The first task was surprisingly tricky given the problem of when to STOP generating Pythagorean Triples, and the second task was reasonably straightforward once I decided how to represent the tree this time to make it easy to traverse "up" to a node's parent, or left or right.


manwar:
Thanks @dcw803

PR #4721, opened 2021-08-16
first submit - working on better solution for ch-1.pl - there is some…

drbaggy:
… pattern in 9^n coming here which I need to get my head round 10^n has 9^n-1 numbers, 2.10^n had 2.9^n numbers, 9.10^n has 8.9^n numbers etc - so should be an easy way to get arbitrary large numbers without a scan!


manwar:
Thanks @drbaggy

PR #4767, opened 2021-08-22
imported my solution's to this week's challenge - 2 nice tasks…

dcw803:
…especially the second one (minesweeper)..


manwar:
Thanks @dcw803

PR #4789, opened 2021-08-24
Add polettix's solution to challenge-127

polettix:

Thank you so much for keeping this going, I think you're the real champion week after week.


manwar:
I love your sweet message, who doesn't like praise anyway. :-D
Thanks @polettix

PR #4813, opened 2021-08-29
jaredor submission for TWC 127

jaredor:
@manwar, in order to catch a step on @gabor you might want to have something to say about this being a "minor" milestone anniversary: 127 == 1111111 in base 2 and next week we'll be entering "Extended ASCII" territory ;-)


manwar:
Nice observation, Thanks for sharing. @jaredor

PR #4828, opened 2021-08-31
Add polettix's solution to challenge-128

polettix:
Hi! Please find my contribution for PWC 128. Still funny after all these weeks, thanks!


manwar:
Thanks @polettix

PR #4845, opened 2021-09-05

dcw803:
I love the second task, which I did as a tiny "discrete event simulation" via building a "diary" of (time,type) pairs - where type is "A" for arrival and "D" for departure.


manwar:
Thanks @dcw803

PR #4869, opened 2021-09-13
Challenege 1

Scimon:
More in a bit.


Scimon:
So the second solution uses my ongoing BinaryTree module https://github.com/Scimon/raku-Tree-Binary

Right now it's a generic Role but I'm about at the point where I'm going to make a search tree that uses the Role and includes code to add and remove nodes.

manwar:
Thanks @Scimon

PR #4879, opened 2021-09-16
Challenge 130 in APL & J

codereport:
Solution in APL & J


manwar:
Welcome on board and thanks for your first guest contributions. @codereport

PR #4894, opened 2021-09-19
imported my solution's to this week's challenges; quite liked the bst question…

dcw803:
(didn't do last week's challenge cos was on holiday, may fill in in sometime later)


manwar:
Thanks @dcw803

PR #4916, opened 2021-09-24
Add polettix's solution to challenge-131

polettix:
Hi! Please find my contributions and my thanks for your hard work :grin:


manwar:
Thanks @polettix

PR #4921, opened 2021-09-25
Document the @xs parameter of challenge 131 task 1

Kaiepi:
Because it needs to be sorted, but the error given otherwise is useless.


manwar:
Thanks @Kaiepi

PR #4930, opened 2021-09-26
Perl solution for challenge 131

adamcrussell:
not much time this week, was able to just do one part


manwar:
Thanks @adamcrussell

PR #4960, opened 2021-10-03
imported my solutions to this week's tasks, including a bonus (ch-2a.pl) which reads the relations from self-describing CSV files.  .…

dcw803:
All good clean fun:-)


manwar:
Thanks @dcw803

PR #4969, opened 2021-10-04
Optimize Ben Davies' challenge 133 task 2 solution

Kaiepi:
`&prime-factors` is eager to invoke `Int.is-prime`; cache an infinite sequence of primes so it gets called once per prime max.


manwar:
Thanks @Kaiepi

PR #4976, opened 2021-10-05
Revise Ben Davies' challenge 133 task 1 solution

Kaiepi:
Actually use one of the square root algorithms suggested.


manwar:
Thanks @Kaiepi

PR #5003, opened 2021-10-10
challenge 133

adamcrussell:
I was experimenting with a Literate Programming framework called nuweb. the nuweb source is a mix of LaTeX and Perl. The generated Perl sources are in the perl directory. Effectively this is just a Perl solution for now, but I will continue experimenting with nuweb in the future I think.


manwar:
Thanks @adamcrussell

PR #5125, opened 2021-10-29
Pwc136

fluca1978:
I had some time to implement possible solutions into *SQL*, in particular PostgreSQL.
I've placed the solutions into the `postgresql` folder, named the solutions as `ch-1.sql` and `ch-2.sql` and added other two blog references as `blog-3.txt` and `blog-4.txt`.


manwar:
Thanks @fluca1978

PR #5126, opened 2021-10-29
Check for even numbers.

Abigail:
In week 136, part 1, if one or both numbers are odd, the pair
cannot be two-friendly.


manwar:
Thanks @Abigail

PR #5131, opened 2021-10-31
Add ch-122 part 1 solution in js

karishmarajput:


manwar:
Thanks @karishmarajput

PR #5156, opened 2021-11-03
Add a "sieve" code to increase the performance of computing lists of Lychrel numbers

drbaggy:
Added optimized version of Lychrel code when used to generate a list of candidate Lychrel numbers - code generates all Lychrel numbers under 1,000,000 in approximately 12 seconds. To 10,000,000 took 6 minutes.


manwar:
Thanks @drbaggy

PR #5175, opened 2021-11-07
imported my solutions to this week's tasks..  task 2 was a bit strange

dcw803:
.. as it's trying to prove a negative


manwar:
Thanks @dcw803

PR #5179, opened 2021-11-08
Task 1 of Challenge 138

ash:
BTW, @manwar, the output for 2021 should be 261, not 260.


manwar:
Opps, sorry. I have updated the task now. Thanks @ash

PR #5189, opened 2021-11-09
Add Python solution to challenge 104

pauloscustodio:


manwar:
Thanks @pauloscustodio

PR #5215, opened 2021-11-14
imported my solutions to this week's questions, including a bonus program

dcw803:
called "tabulate-square-sums" which investigates our split sums more systematically, investigating the first N perfect squares to see which of them have the "split sum is the sqrt" property


manwar:
Thanks @dcw803

PR #5235, opened 2021-11-16
Added challenge 139 solution

Cris-HD:


manwar:
Happy to see you back. Thanks @Cris-HD

PR #5251, opened 2021-11-20
Correct polettix blog entries for PWC139

polettix:
Hi! Sorry for the additional trouble this week, I figured that my blog post paths had to be renamed... so here I am :smile:


manwar:
Thanks @polettix

PR #5256, opened 2021-11-21
139 add jake's solution to challenge 139

dasJake:
hi, here's my first contribution to the weekly challenge ever. i'm really excited.
cheers jake


manwar:
Welcome and thanks for your first contributions. @dasJake

PR #5302, opened 2021-11-29
Challenge 1

Scimon:
Hi, sorry been a bit bleh but back on the challenge. 

I've done a general find N integers with X divisors version which defaults to N=10 and X=8 because... why not? :)


manwar:
Thanks @Scimon

PR #5314, opened 2021-11-30
Add polettix's solution to challenge-141

polettix:
Hi! Early submission this week, so that I'll not forget this time!!! Blog posts are on their way.

Thanks for everything, stay safe with your dear ones 😄


manwar:
Thanks @polettix

PR #5322, opened 2021-12-03
Remove vscode settings from repo

ccntrq:
This folder is in our .gitignore and should not be checked into this
repository. Having this in the repo will affect everyone using vs code
which might be annoying for contributors who don't like the settings.

Me for example ;) I don't like those rulers. :man_shrugging:


manwar:
Thanks @ccntrq

PR #5340, opened 2021-12-06
Solution for week 142.

Abigail:
We lack the resources to implement task 2 (this requires a real time
operating system, and a perl with threads compiled in).

Task 1 is just a repeat of last week, and the week before. I can't be
assed to copy-and-paste most of last weeks code, so, we just have
a Perl solution.


manwar:
Thanks @Abigail

PR #5358, opened 2021-12-10
142

dasJake:
i was discussing with alex (ccntrq) the different definition of "divisor" to last week. at least according to the examples $m itself was not considered a divisor this week. while it makes sense for simplification reasons to exclude an obvious divisor, we wondered if it wasn't sensible to also exclude the '1' in this case?
cheers jake


manwar:
I just made up the rule as go along. There is no rhyme and reason for my decision. I wouldn't claim it to be perfect either. I appreciate your effort. Thanks @dasJake

dasJake:
This is an answer I can live very well with 😄 thanks @manwar

PR #5401, opened 2021-12-21
Add Python solution to challenge 029

pauloscustodio:
Modify example C code to use other that int parameters and results


manwar:
Thanks @pauloscustodio

PR #5433, opened 2021-12-28
Add polettix's solution to challenge-145

polettix:
Hi! It seems that I made it by the end of the (western) year, eventually.

Here is my contribution, as usual.

Let me wish you the best!


manwar:
I wish you Merry Christmas. Thanks @polettix

PR #5441, opened 2021-12-29
Challenges

wlmb:
New solution where I started playing with Object::Pad


manwar:
I might give a try too. Thanks @wlmb

PR #5447, opened 2022-01-02
challenge-145 andrezgz solution

andrezgz:
Happy new year, @manwar! Hope my solution arrives before the due date-time


manwar:
Happy New Year !!!
Thanks for your continuous support. @andrezgz

PR #5450, opened 2022-01-02

dcw803:
however, task 2 is an implementation of the task rather than an implementation of the eertree data structure, because I didn't find the blog clear enough to implement code from..


manwar:
Thanks @dcw803

PR #5451, opened 2022-01-02
palindrome tree cpp code from mohammad khalid anwar

anwarthebest:


manwar:
Welcome on board @anwarthebest

manwar:
FYI, I have renamed the folder to "khalid-anwar" as we already have member with folder "khalid" from Iran.

PR #5463, opened 2022-01-03
Week 146

Abigail:
Part 1 is a fixed output challenge, so, just a glorified Hello World program.
It's stolen from project Euler, task 7.


manwar:
Thanks @Abigail

PR #5471, opened 2022-01-04
Challenge 146 : Back in the saddle

Scimon:
So back on the challenge after the holidays. This weeks were fun and nice and easy in Raku.


manwar:
Thanks @Scimon

PR #5479, opened 2022-01-05
Fixed an error ch-1 notes, added more coding notes to the ch-2 solution explaining the compact version better and the tricks used.

drbaggy:


manwar:
Thanks @drbaggy

PR #5486, opened 2022-01-08
imported my solutions to both tasks

dcw803:
liked both, and only took 30 minutes to do both overall, so nice and straightforward (once you'd worked out the rules for the curious fraction tree)


manwar:
Thanks @dcw803

PR #5519, opened 2022-01-15
Week 147 Task 1 in 8 languages

E7-87-83:
awk
bash
C++
Java
Julia
Node.js
Perl
Smalltalk

with a blogpost


manwar:
Thanks @E7-87-83

PR #5535, opened 2022-01-17
Task 1 - Oops, I thought the challenge was the first 100 Eban numbers... 

PerlBoy1967:
It appears to be Eban <= 100.

Oh well, both are delivered and first 100 is more fun to do efficiently.


manwar:
Thanks @PerlBoy1967

PR #5556, opened 2022-01-23
imported my solutions to this week's puzzles..

dcw803:
I liked both questions, the Eban numbers are very cute.  Cardano triples were initially less pleasant as I dislike floating point algorithms.  But note that I included a bonus ch-2FAST.pl Cardano generator using a sneaky parameterised generation technique.  basic solutions took about an hour to develop, and I spent a second hour researching and build ch-2FAST.


manwar:
Thanks @dcw803

PR #5583, opened 2022-01-30
Added (very late) blog entry for TWC 148.

Util:
manwar, as the originator of the "Cardano Triplets" task, you might be interested in my analysis. Thanks!


manwar:
Thanks @Util

PR #5585, opened 2022-01-30
Fix 149/2 by switching to Math::Base::Convert

choroba:
It can handle larger integers without a problem.


manwar:
Thanks @choroba

PR #5602, opened 2022-02-01
2colours (initial) solutions for challenge No. 150

2colours:
Hello,
I got a bit behind by the overwhelming input size of the last challenge - on the other hand, this week's tasks would run under 1 second even on a soroban. :D 
Thank you for the tasks and see you soon.
M Polgár


manwar:
Thanks @2colours

PR #5619, opened 2022-02-06
imported my solutions to this week's tasks; two nice tasks

dcw803:
two nice questions..  took about 50 minutes overall


manwar:
Thanks @dcw803

PR #5630, opened 2022-02-08
Add polettix's solution to challenge-151

polettix:
Hi Mohammad! Here's my contribution for this week's challenge. Thanks for all the effort and the dedication!


manwar:
Thanks @polettix

PR #5634, opened 2022-02-09
Week 151 Task 2

E7-87-83:
Nice Task 2

The previous pull request overlooked the condition of "always starting with the first".


manwar:
Thanks @E7-87-83

PR #5646, opened 2022-02-13
2colours Raku solutions added

2colours:
Hello,
at last, it's time to add the solutions... I wanted to test them more (especially the first one) but it doesn't seem like I will have enough time for that so let's hope for the best. :)


manwar:
Thanks @2colours

PR #5666, opened 2022-02-17
Solutions from PokGoPun

pokgopun:
Update the codes to generate output in format similar to those shown in examples of the challenge's tasks


manwar:
Did you close by mistake?

PR #5667, opened 2022-02-17
accept accept input array in addition to script's argument, format ou…

pokgopun:
…tput similar to examples of the challenge's task


manwar:
Thanks @pokgopun

PR #5675, opened 2022-02-19
fix bug on parsing cmd args to rectangle's array, revise codes to dyn…

pokgopun:
…amically convert rectangular's array into hash to make it possible in future to support coverage calculation for n rectangles or 3D


manwar:
Thanks @pokgopun

PR #5685, opened 2022-02-20
imported my solutions to this week's tasks, straightforward ones

dcw803:
Had to remind myself of the way of detecting whether two rectangles overlap, and if so what their area is, otherwise straightforward.


manwar:
Thanks @dcw803

PR #5692, opened 2022-02-21
Branch for challenge 153

2colours:
Hello,

it's time to add yet another challenge solution, after some confusion from last week. :)
By the way... may I mention that there is a mistake in the second example of the second task on the website? The input is 125 but then the result takes 123 into account... please have a look at it if you may.
M. Polgár


manwar:
Good catch. Thanks @2colours

PR #5700, opened 2022-02-23
	modified:   challenge-153/mattneleigh/perl/ch-2.pl

mattneleigh:
Minor correction to return value in ch-2


manwar:
Thanks @mattneleigh

PR #5703, opened 2022-02-24
late fix for pwc152 support of n rectangle

pokgopun:
Fix logic that supports n rectangles as well as adding comments to make it easier for later review


manwar:
Thanks @pokgopun

PR #5761, opened 2022-03-10
Add polettix's solution to challenge-155

polettix:
Hi! I thought you did not have much time to read through these comments, but I was clearly wrong 😄 

Please find my contribution for this week's challenge. The wikipedia page for Pisano periods is particularly... *itchy*, so to say, which eventually led me to a brute force attack because, you know... you we asking for the PP modulo 3 😄 

Keep up with the good work and see you next week!


manwar:
Thanks @polettix

PR #5775, opened 2022-03-13
imported my solutions to this week's (rather tricky) challenge

dcw803:
These two tasks were a bit tricky.  My solutions work for the size of problem requested, but I can't guarantee that they always generate the right answers for larger problems, or produce the answers in reasonable amounts of time.


manwar:
Thanks @dcw803

PR #5785, opened 2022-03-16
Solutions for #156

2colours:
Hello,
my latest solutions are here - the rather mundane code doesn't show how much discussion it linked me to, mostly because of junctions that I eventually barely used.
Anyways, this is a part of the process. :)
M Polgár


manwar:
Thanks @2colours

PR #5819, opened 2022-03-23
Week 157 Task 2

E7-87-83:
Finish it faster than I thought; though there were some tricky points for my algorithm (e.g. n=6)


manwar:
Thanks @E7-87-83

PR #5834, opened 2022-03-25
Alternative ch-2 functions added

2colours:
Hello,
thank you for the wholesome tasks, and here goes my weekly solution^^


manwar:
Thanks @2colours

PR #5838, opened 2022-03-27
feat(challenge-157/lubos-kolouch/python/ch-{1,2}.py): Challenge 157 Tasks 1,2 Python LK

LubosKolouch:


manwar:
Thanks @LubosKolouch

PR #5877, opened 2022-04-03
imported my solutions to this week's challenge; 2 nice easy questions

dcw803:
I liked both tasks a lot this week.  Took only 25 minutes in total (reusing my MakePrimes module that I've used in several earlier Perl Weekly Languages).


manwar:
Thanks @dcw803

PR #5878, opened 2022-04-03
Add TWC 158 solutions by Bruce Gray : Raku, Perl, Go.

Util:
Raku and Perl also have .sh files containing additional one-liner versions.


manwar:
Thanks @Util

PR #5895, opened 2022-04-06
Add Perl solution to challenge 040

pauloscustodio:


manwar:
Thanks for your ongoing support, much appreciated. @pauloscustodio

PR #5899, opened 2022-04-07
Fix the directory name

choroba:
We don't use "perl5" anymore. It even causes problems reporting the
solutions as being in an "additional language".


manwar:
Thanks @choroba

PR #5910, opened 2022-04-10
2 nice tasks this week, took 1h15m for both basic solutions - then another 10 minutes to add a bonus feature..

dcw803:
bonus feature was (in task 2) a "--tabulate" mode which makes it display the first N Moebius numbers in tabular form. ./ch-2 -t 50 generates the same table as the Wikipedia page contains..


manwar:
Thanks @dcw803

PR #5919, opened 2022-04-11
Challenge 1 and 2

Scimon:
Note for challenge 2 the script outputs all the equilibrium points not just the first. 

A fun test list 1, -1, 1, -1, 1 (keep adding -1, 1 pairs as long as you like) every point is an equilibrium point.


manwar:
Thanks @Scimon

PR #5924, opened 2022-04-11
	modified:   challenge-160/mattneleigh/perl/ch-1.pl

mattneleigh:
Minor correction to output in ch-1.pl


manwar:
Thanks @mattneleigh

PR #5934, opened 2022-04-14
Add polettix's solution to challenge-160

polettix:
Hi! I hope you're doing great, thanks for your amazing dedication.


manwar:
THanks @polettix

PR #5942, opened 2022-04-15
feat(challenge-160/lubos-kolouch/p[erl,ython]/ch-[12].p[ly]): Challenge 160 Task 1 2 LK Perl Python

LubosKolouch:


manwar:
Thanks @LubosKolouch

PR #5953, opened 2022-04-17
imported my solutions to this week's tasks; quite easy and fun tasks

dcw803:
took approx 40 mins all told


manwar:
Thanks @dcw803

PR #5969, opened 2022-04-20
Perl solution to 161

simbabque:


manwar:
Welcome back and thanks for your contributions. @simbabque

PR #5996, opened 2022-04-24
Solutions for week #161

2colours:
Hi,
sorry for the inconvenience, I noticed something that could definitely be improved (the optimally short solution doesn't necessarily have monotonously increasing length for words).


manwar:
Thanks @2colours

PR #5997, opened 2022-04-24
imported my solutions to this week's tasks - both Perl

dcw803:
and ch-1 done in C as a bonus.  Took approx 1 hour for the Perl ones, and another hour for the C version of ch-1


manwar:
Thanks @dcw803

PR #5998, opened 2022-04-25
Partial challenge 161 by Jaldhar H. Vyas.

jaldhar:
Perl version of problem 2 still has me tied in knots so I'm going to have to spend more time on that one and on the blog post but I have done problem 1 and the raku solution of problem 2.


manwar:
Thanks @jaldhar

PR #6037, opened 2022-05-01
Huge suffering @ 162nd week

2colours:
Hello,
The suffering was real as I finished the second task; at least I've come across loads of novelties. :) I hope at least I didn't mess up what is supposed to just work as intended.
Thanks for the challenges as usual. :+1:


manwar:
Thanks @2colours

PR #6038, opened 2022-05-01
imported my solutions to this week's 2 tasks

dcw803:
(plus a bonus ch-1 in C in C/ch-1.c).. two enjoyable tasks which took 1h20m in total.  unfortunately I only left 1h10m before the deadline:-)


manwar:
Thanks @dcw803

PR #6039, opened 2022-05-02
fix double letter handling

Hiranyaloka:
Oops, found a bug in the double letter handling. Maybe too late?


manwar:
Thanks @Hiranyaloka

PR #6053, opened 2022-05-03
Weekly solutions (2colours)

2colours:
Hello,
this week seems to be more chill, the second task was still very educational for me but at least I'm not lagging behind this week. :D So thank you, as usual, and please take my humble weekly attempt :P


manwar:
Thanks @2colours

PR #6057, opened 2022-05-04
Add polettix's solution to challenge-163

polettix:
Hi! Thank you so much for last week, I'm ashamed to have forgotten about publishing my solution here. Thanks again!


manwar:
Thanks @polettix

PR #6067, opened 2022-05-07
Challenge 163 solution and some old challenges

ccntrq:
My solutions for the current challenge and few solutions for old challenge in haskell


manwar:
Thanks @ccntrq

PR #6069, opened 2022-05-08
Week 163 from CY

E7-87-83:
2 Perl scripts

For Task 2 Guest contribution
1 Python script
1 Julia script


manwar:
Thanks @E7-87-83

PR #6074, opened 2022-05-08

dcw803:
Done both in perl, C and pascal this time. Two nice problems.


manwar:
Thanks @dcw803

PR #6100, opened 2022-05-14
Add polettix's solution to challenge-164

polettix:
Hi! Back on track, I hope... thanks again for last week. Cheers!


manwar:
Thanks @polettix

PR #6111, opened 2022-05-15

dcw803:
easy tasks this week, took roughly 30 minutes to do both tasks in Perl, another hour or so to translate to C and get working quickly.


manwar:
Thanks @dcw803

PR #6131, opened 2022-05-20
Week 165

wg-romank:
Hey,

Thank for putting together the challenge!

I've played around with it in J, adding some scripts for fitting linear regression.
It also includes the resulting `svg` in scaled axes, not sure if it's a good idea to put it there along with input since it's few hundred kb, if not please let me know I'll drop those files.

Cheers,
-Roman


manwar:
Thanks @wg-romank

PR #6141, opened 2022-05-22
imported my solutions to the SVG tasks, in C and Perl

dcw803:
(gotta get those guest contribs above Colin:-))


manwar:
Thanks @dcw803

PR #6177, opened 2022-05-29
Challenge 166

simbabque:
All of this code is written with Github Copilot. Blog is coming.


manwar:
Bonus blog this time, wow. Thanks @simbabque

PR #6204, opened 2022-06-05
Add polettix's solution to challenge-167

polettix:
Hi! Sorry for burdening you last week, I had a few personal troubles (solved) and that completely passed under the radar!

Keep up with your great work, your dedication is an inspiration for me and many others!


manwar:
Thanks @polettix

PR #6206, opened 2022-06-05
imported my solution to task 1 of this week's tasks

dcw803:
although I've decided not to do task 2 as it's way too mathematical for me.  Complex numbers, the Gamma function that I've never heard of, and formulae that sound way too complex to understand: that's not for me.

Also, my task 1 gives more circular primes than your output, i.e. all of them, because that's what the question asked for, whereas I suspect the output shown is not actually "the first 10 N-or-more digit circular primes", but "the first 10 N-or-more digit smallest representative prime from all known cycles of circular primes", and that's not what the question asked for.

### DIFF
--- a/challenge-167/duncan-c-white/README
+++ b/challenge-167/duncan-c-white/README
@@ -1,75 +1,51 @@
-TASK 1: Hexadecimal Words
+TASK 1: Circular Prime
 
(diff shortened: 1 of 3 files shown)


manwar:
Thanks @dcw803

PR #6227, opened 2022-06-08
pwc145 solution in go - ch-2 to handle the case where a letter happen…

pokgopun:
…s more than twice in a word


manwar:
Thanks @pokgopun

PR #6231, opened 2022-06-09
Answer to challenge 168 by Massa, Humberto

massa:
There was one little thing that I didn't like in the challenge 1: I had to guesstimate how many Perrin primes would be enough to have 13 distinct primes, went with 20, and was successful with not a lot of overhead.


manwar:
Thanks @massa

PR #6256, opened 2022-06-13
Small typo.

jaldhar:
I accidently left in a redundant line from an older version of my Perl solution to task 1.


manwar:
Thanks @jaldhar

PR #6269, opened 2022-06-15
Raku: Achilles numbers

habere-et-dispertire:


manwar:
Thanks @habere-et-dispertire

PR #6282, opened 2022-06-18
Add polettix's solution to challenge-169

polettix:
Here we go with this week's contribution. Thank you for running this and guarding our backs (like the two or three times you did the work to publish my solutions because I forgot...).


manwar:
Thanks @polettix

PR #6290, opened 2022-06-19
Weekly solutions by 2colours

2colours:
Hello,
it's better late than never, so I put together an attempt for this weekly again. Thank you for making us think this week as well. :)


manwar:
Thanks @2colours

PR #6296, opened 2022-06-19
imported solutions tp this week's tasks

dcw803:
two quite nice tasks - but I'm getting very bored of prime-based number theory tasks.


manwar:
Thanks @dcw803

PR #6319, opened 2022-06-22
one-liner with regex for challenge 170

brxfork:
Some fun with regex engine.


manwar:
Welcome on board and thanks for your first contribution. @brxfork

PR #6324, opened 2022-06-23
one-liner with regex for challenge 144

brxfork:
Too late for old challenges?


manwar:
Not at all, please feel free. Thanks for your contributions. @brxfork

PR #6328, opened 2022-06-24
sgreen solutions to challenge 170

simongreen-net:
Not sure if I'll have time to write a blog, I've got a five hour flight tomorrow (holiday!), so might do it then.


manwar:
Thanks @simongreen-net

PR #6345, opened 2022-06-26
imported my two perl solutions (45 mins), then went back and translated them to C (another hour)

dcw803:
two nice easy tasks this week (I'm not complaining:-))


manwar:
Thanks @dcw803

PR #6352, opened 2022-06-27
Abundant use of regex engine to find 20 Abundant Odd Numbers

brxfork:
Fun with regex engine to find OEIS A005231


manwar:
Thanks @brxfork

brxfork:
Thank you very much @steve-g-lynn

PR #6381, opened 2022-07-03
Add polettix's solution to challenge-171

polettix:
Hi! Better near-to-late than never! Thank you for all the hard work and keep it up, you're the best!


manwar:
Thanks @polettix

PR #6382, opened 2022-07-03
imported my solutions to this week's tasks - 2 in Perl and 1 in C..

dcw803:
really liked both of these tasks, especially the higher order function one..


manwar:
Thanks @dcw803

PR #6389, opened 2022-07-04
pwc172 solution in go

pokgopun:


manwar:
Can you please share your email address so that prize money can be send it to you?
You can email us to perlweeklychallenge@yahoo.com.
Thanks @pokgopun

PR #6400, opened 2022-07-06
Add polettix's solution to challenge-172

polettix:
Hi! Please find my contribution for this week's challenge. Thanks for the dedication and the infinite well of interesting topics!


manwar:
Thanks @polettix

PR #6426, opened 2022-07-11
Challenge 173

Scimon:
Moved into new house things calming down, figured I'd give this week a shot (also both are pretty easy in the Raku's)


manwar:
Thanks @Scimon

PR #6439, opened 2022-07-13
Add polettix's solution to challenge-173

polettix:
Hi! Here it is my contribution for this week's challenge. I hope my questions don't bother you, anyway no real questions this time in the second challenge... and for the first too, actually :grin:


manwar:
Thanks @polettix

PR #6494, opened 2022-07-24
solutions to ch#'174 plus belated attempts at ch#172 and 173

dcw803:
Dunno if I'm too late to submit 172 and 173, I felt over-hot and tired the last couple of weeks, and skipped doing them.. This is a catchup commit comprising:

- solutions to ch#172 in Perl and C
- solutions to ch#173 in Perl and C
- solutions to ch#174 in Perl (with some optimizations of ch-1), and ch-1 in C as well (with one more optimization)


manwar:
Thanks @dcw803

PR #6516, opened 2022-07-29
Weekly solutions (175) by 2colours

2colours:
Hello,
those 20-digit numbers last week got a little scary (and hopeless to execute before the end of the universe) but this week, even the perfect totients could execute within a couple of seconds so I was content with the result. Please take my weekly solutions.


manwar:
Thanks @2colours

PR #6524, opened 2022-07-30
Solution to Challenge 2 in Perl and Go

Bhat-Gurunandan:
Hi Mohammad, 

An interesting thing happened when I started to code the solution to Challenge-2. I was playing with the code and trying different variations when I discovered that if you increment the iterated sum of totients by 1 (add the successive totients and then add 1) and then you check which number matches the sum, you get perfect powers of 2. (2, 4, 8, 16, ....)!!! I spent a lot of time looking at Number Theory sources to find if this was discussed somewhere. Sadly could not find any. Maybe someone on the list who is strong on Number Theory might explain why that happens. :)

In any case, the bad news is that I spent so much time trying to investigate this startling (to me, maybe it is commonplace for number theorists) that I have no time to code the solution to Challenge-2. I am aware that Lewis Carroll (he of Alice fame) has an algorithm to do that and I always wanted to code it but sadly I have a "No Coding Sunday" policy at home, so will have to attempt this only next week. I hope you do not mind.  Sincere apologies and have a great weekend. Cheers

### DIFF
--- a/challenge-175/bhat_gurunandan/go/ch-2.go
+++ b/challenge-175/bhat_gurunandan/go/ch-2.go
@@ -1,0 +1,72 @@
+package main
+
+import (
+	"fmt"
+	"log"
+	"os"
+	"strconv"
+)
+
+func gcd(n int, m int) int {
+
+	for m != 0 {
+		n, m = m, n%m
+	}
+
+	return n
+}
+
+func totient(n int) int {
+
+	totient := 1
+
+	for this := n; this > 1; this-- {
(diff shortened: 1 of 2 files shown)


Bhat-Gurunandan:
I just noticed that I have submitted code with this incremented by 1 solution so I have to send you one more pull request. Clearly, I am not there and my brain is numb with numbers ;) Requesting shortly.

Bhat-Gurunandan:
So added the commit. Feel free to merge at your convenience.

manwar:
Thanks @Bhat-Gurunandan

PR #6615, opened 2022-08-14
Perl solution for challenge 2

aut0exec:


manwar:
Welcome on board and thanks for your first contribution. @aut0exec

PR #6623, opened 2022-08-16
Add polettix's solution to challenge-178

polettix:

I hope you completely recovered :-)


manwar:
Thanks, neck is lot better. I don't need neck guard anymore. @polettix

PR #6647, opened 2022-08-23
belatedly had a crack at challenge 177 (was on holiday at the time)

dcw803:
… 2 nice tasks, done in C and Perl


manwar:
Thanks @dcw803

PR #6662, opened 2022-08-28
imported my belated attempt at challenge 178, still away last week..

dcw803:
base -2i is very weird, the business date task was quite interesting to do without pre-built business hours support.  Translated both tasks to C as well.


manwar:
Thanks @dcw803

PR #6665, opened 2022-08-29
Weekly solutions by 2colours

2colours:
Hello,
after a lengthy break, I'm back with these two cute little nuances.
I have some incomplete solutions from the last couple of weeks, especially the quater-imaginary stuff is something I really want to finish, so I might be back with some post-due solutions as well, later. :)


manwar:
Welcome back and thanks @2colours

PR #6677, opened 2022-08-30
my quick-and-dirty raku answers (to ch-180), as usual

massa:
They could both benefit from longer explanations, but raku made both operations quite trivial...

### DIFF
--- a/challenge-180/massa/raku/ch-1.raku
+++ b/challenge-180/massa/raku/ch-1.raku
@@ -1,5 +1,4 @@
 use v6;
 
 unit sub MAIN(*@args);
-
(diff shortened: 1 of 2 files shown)


massa:
Please, don't merge this yet, I found out it's still not right -- it passes the challenge's strings, but fails for
`"my quick-and-dirty quack same saney raku answers (to ch-180), as usual #6677"`
-- it gives `13`, for the positition of the second `d`, which is obviously wrong...

massa:
And now I'm satisfied! It works! 😄

manwar:
Thanks @massa

PR #6684, opened 2022-09-02
Nothing like implementing in another language...

massa:
to find a bug in your implementation :-)

please merge; I found a "reuse of the `%visited` hash" bug when I compared the output with the (included) C++20 version.


manwar:
Thanks @massa

PR #6690, opened 2022-09-04
Challenge 180 in Clojure and Fennel

tylerw:
Hello @manwar, 

I believe this is the first contribution in [Fennel](https://fennel-lang.org/)!


manwar:
Thanks @tylerw

PR #6727, opened 2022-09-11
Branch for challenge 181

2colours:
Hello,

last but maybe not least, I made it with the weeklies. The first solution is very optimistic but I really wanted to "make a debut" with Raku's grammars and I didn't have a lot of time to forge a more comprehensive program... Anyway, please take this humble attempt and keep up the good work with the challenges, definitely learning a lot on the way still. ^^


manwar:
Thanks @2colours

PR #6728, opened 2022-09-11
imported my solutions (Perl only this week) for this week's tasks.

dcw803:
C versions might follow tomorrow.  two nice tasks.


manwar:
Thanks @dcw803

PR #6737, opened 2022-09-13
bugfix plus challenge 181 tasks in C..

dcw803:
can't have C++ beating C in the guest language stakes:-)


manwar:
It shows your love for C. Thanks @dcw803

PR #6739, opened 2022-09-14
Guest contribution / challenge 182 / happy-barney

happy-barney:
PR also contains poc for shared tests (should used naming pattern be required and calling convention specified)


manwar:
Welcome on board and thanks for your first contributions. @happy-barney

PR #6781, opened 2022-09-23
Add polettix's solution to challenge-183

polettix:

I skipped Raku in the second task because... I could not come up with a nice compact way to do things, I hate fiddling with dates and to be honest this years+days thing made little sense to me :grin:


manwar:
Thanks @polettix

PR #6782, opened 2022-09-24
Branch for challenge 183

2colours:
Hi,

thank you for the challenges. I personally like that they are often a bit vague because this leaves some design decisions for the solver. In this case, I spent more time deciding on the input and the output, also trying to conform with the given examples:

- ch-1: accepting JSON-compatible content, inside parentheses for the containing list. The output is also provided as JSON arrays inside a list (utilizing two Raku modules that can be installed via zef)
- ch-2: reverse dates produce negative result, 0 is not shown - TODO handle matching dates - there is no conversion between days and years so that leap-years don't cause problems, instead, the year is offset to the closest date within the interval

Please take my solution (after a little modification on my side, noticed the matching date situation). :)

### DIFF
--- a/challenge-183/2colours/raku/ch-1.raku
+++ b/challenge-183/2colours/raku/ch-1.raku
@@ -15,8 +15,8 @@ my \try-list = try $input.&from-json;
 when not try-list.defined {
   say 'The elements of the list should be valid according to JSON.';
 }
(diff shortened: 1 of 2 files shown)


manwar:
Thanks @2colours

PR #6792, opened 2022-09-25
imported my solutions to both tasks in Perl, and to task 1 in C

dcw803:
didn't have time to build a C solution too using mktime(), may do that later on.


manwar:
Thanks @dcw803

PR #6829, opened 2022-10-02
imported my solutions (both tasks in Perl and C) to this week's tasks

dcw803:
two nice easy tasks this week.. good fun


manwar:
Thanks @dcw803

PR #6830, opened 2022-10-02
Weeklies by 2colours

2colours:
Hello,

a mini solution written in a hurry but I didn't want to skip this week. :D 
I think it's easy to see by the solution but it's worth clarifying nevertheless: both solutions have optimistic input handling - in the first case, data is consumed in [a-z]{2}[0-9]{4} blocks, in the second case, it's consumed between apostrophes and then character by character, optionally space-separated.

Looking forward to next week, and please excuse me for the superficial solutions. ^^


manwar:
Thanks @2colours

PR #6845, opened 2022-10-04
First pass - lots of one-liner comparisons

drbaggy:


manwar:
Cool creations, thanks @drbaggy

PR #6848, opened 2022-10-04
Challenge 185 - guest contribution happy-barney

happy-barney:


happy-barney:
and next contribution may contain generic perl tester with per-challenge data (should contribution contain predefined function)

PR #6851, opened 2022-10-04
Guest contribution - happy-barney - alternative mask code

happy-barney:
with shared, challenge independent, test script and related test data


manwar:
Thanks @happy-barney

PR #6858, opened 2022-10-06
Add polettix's solution to challenge-185

polettix:
Hi! Early contribution, just to balance last week. The second blog post is on its way, due tomorrow. Cheers and thanks for all the hard work and dedication!


manwar:
Thanks @polettix

PR #6866, opened 2022-10-08
Add solution by karishma rajput

karishmarajput:
solution of ch-001 in js


manwar:
Thanks @karishmarajput

PR #6867, opened 2022-10-08
Raku beating challenge 185 :P

2colours:
Hello,

thank you for the weeklies, although this week Raku clearly did the heavy lifting for me. Anyway, sometimes it's good to have it easy; more time to focus on other things, like trying to lift Raku itself. :D


manwar:
Thanks @2colours

PR #6868, opened 2022-10-08
Solved week 185 challenge 1in APL

JoaoFelipe3:


manwar:
Welcome on board and thanks for your first guest contribution. @JoaoFelipe3

PR #6872, opened 2022-10-09
feat(challenge-185/lubos-kolouch/python/ch-{1,2}.py): Challenge 185 Task 1 2 LK Python

LubosKolouch:


manwar:
Thanks @LubosKolouch

PR #6909, opened 2022-10-15
Week 186 Codes

E7-87-83:
2 Perl scripts, 1 Java soln for Task 2
I am not bored by the Task 2 and writing a blogpost for it!


manwar:
Thanks @E7-87-83

PR #6912, opened 2022-10-16
Raku ft. 2colours solved challenge 186

2colours:
Hello,

another chill week regarding weeklies as Raku solves the tasks by itself. :P I'm hoping for some arithmetics to return to the weeklies. Anyways, please take this really humble and minimalistic submission, trying to be persistent. :)


manwar:
Thanks @2colours

PR #6917, opened 2022-10-16
imported my solution - only to task 1 - this week, in Perl and C

dcw803:
I didn't do task 2 cos I really hate Unicode, never learned how to do it in Perl, and therefore have no clue how to solve it


manwar:
Thanks @dcw803

PR #6949, opened 2022-10-23
imported my solutions to both tasks in C and Perl

dcw803:
two nice tasks, although I did task 2 in "blunderbuss mode", I wondered if there was a more elegant way that I was missing


manwar:
Thanks @dcw803

PR #6984, opened 2022-10-30
Weekly challenges by 2colours

2colours:
Hello,

I still haven't gotten around to do the next weekly but today I've done so much Raku that I had inspiration to do this quickly. :D
Hopefully the code is simple enough to understand for someone who knows Raku - although the expressivity of this regex sublanguage is still surprising for me as well...
The other thing is, I felt the second challenge was a bit hard to decipher. The examples suggest that $x == $y > 0 turns into $x == $y == 0 in one step, that's the presumption I used.

Thank you for the persistence, and I hope the things I've been working on, instead of the last weekly, will start paying out for Raku as well. :P


manwar:
Thanks @2colours

PR #6994, opened 2022-10-30
imported my solutions to this week's tasks…

dcw803:
… two nice and easy ones this week, both done in Perl and C


manwar:
Thanks @dcw803

PR #7015, opened 2022-10-31
Week 189 - solution for @spazm in perl 5

spazm:
Ch2 was tricky for quite a bit.

The key was realizing that any minimum length set would span from the first to last of the element setting the degree.


spazm:
ObComment: first!

PR #7017, opened 2022-11-01
Current weeklies by 2colours

2colours:
Hi,

thanks for the challenges - they help me develop/revise my coding conventions. My latest favorite input processing is: I take the format of the provided examples and parse CLI arguments based on that format. I'm not saying that this is a superior style but a good way to practice regex parsing and also an approach that can be applied consistently.
Keep up the good work - I hope I can, too. :)


manwar:
Thanks @2colours

PR #7054, opened 2022-11-07
Weeklies no. 190 by 2colours

2colours:
Hello,

some of the solutions might be a little on the optimistic end but it was fun and also, I'm not sure if I'd have time later on, so there it goes...
Thank you for keeping up the good work and please take my solutions-in-a-hurry. :)


manwar:
Thanks @2colours

PR #7072, opened 2022-11-13
Robbie hatley 190

LoneWolfiNTj:
#1 was easy; but #2 took a while. :-)


LoneWolfiNTj:
The "Correcting Errors" commit is erroneous. That was my botched attempt at correcting discrepancies between the PWCC fork on my computer and the PWCC fork on my Github account. It should never have been forwarded upstream (I'm not sure why that happened), and it was done incorrectly anyway. Please delete.

PR #7086, opened 2022-11-14
Both challenges.

Scimon:
Note challenge 1 returns True / False 

Challenge 2 will calculate all values between 1 and 15 if not given a value (and in the same time (near enough) it takes to do 15 by itself).


manwar:
Thanks @Scimon

PR #7097, opened 2022-11-17
Adding a second way of doing it and benchmarking code

Scimon:
Adds a second way of working out the answer and benchmarking for the two.

Turns out, about the same speed. 

Also adds tests for the single and empty list cases.


manwar:
Thanks @Scimon

PR #7102, opened 2022-11-18
Add polettix's solution to challenge-191

polettix:
Hi! Thanks for your dedication... and your patience with my questions :grin:


manwar:
Thanks @polettix

PR #7120, opened 2022-11-20
Add TWC 191 solutions (Raku only) by Bruce Gray.

Util:
Second task references work by @jo-37 , @2colours , and @abigail .


manwar:
Thanks @Util

PR #7123, opened 2022-11-20
imported my solutions to this week's tasks in C and Perl, plus ch-1 from last week in Pascal

dcw803:
two nice tasks this week. although I'm not absolutely sure I've got cute list quite right, a few more examples would have helped.  I've tabulated my results (number cute lists of lenth n) for n=1..10..


manwar:
Thanks @dcw803

PR #7143, opened 2022-11-22
refactored Python solution

ealvar3z:
Per the FB thread between @James_Smith and me. I refactored my Python
code to handle bigger inputs and optimize the code.


manwar:
Thanks @ealvar3z

PR #7153, opened 2022-11-26
A shot at week 192 by 2colours

2colours:
Hello,

thank you for the interesting challenges, I learned from them again... I can only hope my heuristics is both correct and well-implemented for the second task but I didn't feel I could prove it and I didn't have more time for it now. :D 

Anyways, off we go - please let them make the the collection of solutions a tad more colorful. :P


manwar:
Thanks @2colours

PR #7164, opened 2022-11-27
Add polettix's solution to challenge-192

polettix:
Hi! Please find this week's contribution, and thanks for the great work!


manwar:
Thanks @polettix

PR #7212, opened 2022-12-05
slight tweak to ch-2.pl and then translated it into C…

dcw803:
(see C/ch-2.c for that), sorting an array of pairs (strings + difference signatures) by signature using qsort, and then walking the sorted array of pairs looking for the signature to change, i.e. edges.


manwar:
Thanks @dcw803

PR #7242, opened 2022-12-12
imported my (slightly late) solutions to challenge 194, both in Perl and C

dcw803:
two nice easy questions which I enjoyed


manwar:
Thanks @dcw803

PR #7267, opened 2022-12-18
Current weeklies by 2colours

2colours:
Hello,

Thank you for the unusual pair of tasks - this time, the first one seemed much more complex. Perhaps I could have polished it further but I kinda just rolled with it once it seemed to return legitimate values...
After a couple of busy weeks, these challenges seemed to be a good start to get something done for AoC as well. :P Please take these solutions with absolutely no warranty.


manwar:
Thanks @2colours

PR #7274, opened 2022-12-18
added my solutions to this week's 2 tasks, both nice and easy,

dcw803:
did both in Perl and then in C

(ch-2.c needed to simulate a frequency hash of arbitrary integers - which I did as a list of pairs)


manwar:
Thanks @dcw803

PR #7289, opened 2022-12-20
feat: solution for challenge 196

carlos157oliveira:


manwar:
Thanks @carlos157oliveira

PR #7381, opened 2023-01-08
committed my solutions to this week's challenge, in Perl and C

dcw803:
skipped last week's as a NY break, but will do them soon..  happy new year


manwar:
Happy New Year !!!
Thanks @dcw803

PR #7435, opened 2023-01-18
Add polettix's solution to challenge-200

polettix:
Hi! Thank you for all the dedication and for your help!


manwar:
Thanks @polettix

PR #7453, opened 2023-01-23
Partial solution of week 200 by 2colours

2colours:
Hello,

Thank you for the tasks, interesting week again.
I was too late with this solution, also I had to realize that if I cannot start with ch-2 the way I'd like to because that would take so much time inventing some ultimate terminal display logic that I would be completely drifting away. :D Also, I used some things in ch-1 that I don't even fully understand why they work... but they seem to work so I hope we can spot the mistakes on the next raku  study group meeting, should there be some.
Cheers :P


manwar:
Thanks @2colours

PR #7466, opened 2023-01-24
Prolog solutions for week 201 by 2colours

2colours:
Hello,

this time, by the nature of the problems, it really got tempting to try and refresh my Prolog memories. I don't aim for perfection but I was happy to get to a point where I started getting legitimate results - also, performance considerations are always interesting with this language.
I don't know yet if I'm gonna also send Raku solutions this week - for now, please take this humble nostalgic attempt. :)


manwar:
Thanks @2colours

PR #7505, opened 2023-01-30
Add solutions to 202: Consecutive Odds & Widest Valley by E. Choroba

choroba:
202/2 also solved in Erlang.


manwar:
Thanks @choroba

PR #7530, opened 2023-02-06
Weeklies by 2colours

2colours:
Hello,

sorry for being a bit late - the more interesting task really took its prey. :) I hope I nailed it because that one involved a lot of planning and reasoning, rather than clean self-explanatory code...
Anyway, here we are. Looking forward to next week, just like the solutions of Raku Study Group fellas. :D


manwar:
Thanks @2colours

PR #7533, opened 2023-02-06
(belatedly did and) committed my solutions to this challenge

dcw803:
did both tasks, in C and Perl as usual; I thought the widest valley was going to be trickier, but it seemed easy and worked first time


manwar:
Thanks @dcw803

PR #7541, opened 2023-02-08
solution to challenge 203

carlos157oliveira:
Good evening manager of Perl Weekly Challenge. Where do I send a profile picture?


manwar:
Thanks @carlos157oliveira 

Please send the profile photo to perlweeklychallenge@yahoo.com

PR #7544, opened 2023-02-09
Challenge 203.1: Solution in SWI Prolog

shawnw:
Been a while since I've done one of these, so decided to do something different.


manwar:
Thanks @shawnw

PR #7574, opened 2023-02-14
Improved monotonic-determination function.

robbie-hatley:


manwar:
Thanks @robbie-hatley

PR #7598, opened 2023-02-20

dcw803:
I slightly cheated with the C resize-matrix task: didn't actually build a 2d matrix of the right size and print that out - simply printed out the 1d array of elements as if it were an RxC matrix..


manwar:
Thanks @dcw803

PR #7637, opened 2023-02-26
Add TWC 205 blog post and solutions by Bruce Gray (Perl and Raku).

Util:
@manwar, as the author of the Max XOR problem, you may be interested in my blog post, which details a fast solution when the input array is large.

Also, I am blocked on the interview, trying to resolve. Thanks!


manwar:
Thanks @Util

PR #7643, opened 2023-02-27
A trick?

drbaggy:


carlos-gentrop:
Hi. Good afternoon. I ended up reading your solutions before doing my own because your pull request title called my attention (I wasn't expecting that it would name your solutions). I will try to make different, and it will be, because what you've done is very original. I fancy the trick you discovered for the second task, thanks for showing. I liked your code for the first task but I stumbled on some interpretation problems. For example this line `abs( $t+$_-1_440 ) < $min && ( $min = abs $t+$_-1_440 ) for @_;` I was expecting to give the distance by going counter-clockwise, but I tested with 23:59 (1439 min) and 00:30 (30 min) and the formula gave 1439 + 30 -1440 = 29 min, while I was expecting 31 min. Did I misinterpret what this line means?

carlos157oliveira:
Hi. Good afternoon. I ended up reading your solutions before doing my own because your pull request title called my attention (I wasn't expecting that it would name your solutions). I will try to make different, and it will be, because what you've done is very original. I fancy the trick you discovered for the second task, thanks for showing. I liked your code for the first task but I stumbled on some interpretation problems. For example this line `abs( $t+$_-1_440 ) < $min && ( $min = abs $t+$_-1_440 ) for @_;` I was expecting to give the distance by going counter-clockwise, but I tested with 23:59 (1439 min) and 00:30 (30 min) and the formula gave 1439 + 30 -1440 = 29 min, while I was expecting 31 min. Did I misinterpret what this line means?

manwar:
Thanks @drbaggy

PR #7647, opened 2023-02-27
Weeklies for 206th week by 2colours

2colours:
Hello,

thank you for the challenges - including the last couple of weeks, I was really cracking my head, waiting for the right solutions to come. :D After a couple of swingy weeks, I felt obliged to make this return, and I needed more reasoning with the second task than I'd like to admit, so the apparently simply solutions are a bit misleading, that was still kind of challenging to convince myself.
Please take these solutions; hopefully I got them right but if not, it will probably be clarified on this week's Raku Study Group online meetup. :)

### DIFF
--- a/challenge-206/2colours/raku/ch-1.raku
+++ b/challenge-206/2colours/raku/ch-1.raku
@@ -1,0 +1,26 @@
+#!/usr/bin/env raku
+
+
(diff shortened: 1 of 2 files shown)


manwar:
Thanks @2colours

PR #7663, opened 2023-03-05
Turns out, with hints from others, we don't need to brute force here

kjetillll:
...and loop through every possible pairings permutation.

Just need to sort the input numbers and sum up every evenly indexed (0-indexed array) number.


manwar:
Thanks @kjetillll

PR #7674, opened 2023-03-05
catchup commit: challenge 205, both tasks, Perl and C, plus challenge 206: both in perl only…

dcw803:
had a lousy cold last week, so ignored PWC last week..  may consider doing challenge 206 tasks in C later


manwar:
Thanks @dcw803

PR #7690, opened 2023-03-07
ok, a day late but I've done the C versions of both tasks in ch 206…

dcw803:
…(and added a bit more debugging to ch-2.pl too)


manwar:
Thanks @dcw803

PR #7720, opened 2023-03-12
imported my solutions to both tasks, in Perl and C as usual

dcw803:
two nice simple tasks this week.  nice!


manwar:
Thanks @dcw803

PR #7760, opened 2023-03-19

dcw803:
two nice tasks. although for some reason I had difficulty thinking task 2 out clearly


manwar:
Thanks @dcw803

PR #7805, opened 2023-03-26
Weeklies by 2colours

2colours:
Hello,

thank you for the weeklies - very interesting things came up while working on the solution. Unfortunately, I don't have more time to polish it so I kind of just hope I got the second task right.
Take care.


manwar:
Thanks @2colours

PR #7807, opened 2023-03-26
imported my solutions to this week's tasks, task 1 in Perl and C, task 2 in Perl…

dcw803:
..will do task 2 in C tomorrow when I have time


manwar:
Thanks @dcw803

PR #7817, opened 2023-03-28
belatedly added my C implementation of task 2

dcw803:
really quite tricky to get right, had to build a simple set module (slist.[ch]) to do it. memory management was especially tricky.


manwar:
Thanks @dcw803

PR #7885, opened 2023-04-11
Solve PWC212

wlmb:
I believe there was a mistake in the statement of task 2, in example 3. I guess the correct answer should be -1 unless the size is changed to 3. Regards.


manwar:
> I believe there was a mistake in the statement of task 2, in example 3. I guess the correct answer should be -1 unless the size is changed to 3. Regards.

Yes, you are right. I have fixed the typo. Sorry about that.

manwar:
Thanks @wlmb

PR #7923, opened 2023-04-17
imported my C solutions to challenge 212 a day late

dcw803:
also tweaked some comments in the top level README file and ch-2.pl


manwar:
Thanks @dcw803

PR #7941, opened 2023-04-22
213th week by 2colours

2colours:
Hi,

thank you very much for the challenges; it took me some time to return, apparently misunderstood last week's tasks, for example. This time, something usual happened: I hope that the solution is alright but Raku itself caused me some trouble so I won't feel guilty if something went wrong. :P
Please take these quirky solutions nevertheless.


manwar:
Thanks @2colours

PR #7981, opened 2023-04-28
belatedly finished and committed last week's task 2 in C (C/ch-2.c)

dcw803:
seem to have got it working fine


manwar:
Thanks @dcw803

PR #7993, opened 2023-04-30
imported my solutions to this week's tasks (both in Perl, first in C)

dcw803:
.. will do 2nd task in C but not tonight as I'm tired..


manwar:
Thanks @dcw803

PR #8061, opened 2023-05-12
Solutions to challenge 216

jo-37:
GLPK seems to be the perfect tool to solve task 2.  Sadly, its perl integration is outdated. So I had to learn a new language to use it. Here is my first program written in octave.


manwar:
Thanks @jo-37

PR #8099, opened 2023-05-18
Robbie Hatley's solutions in Perl for The Weekly Challenge 217

robbie-hatley:
Instead of using a "special sort" approach for #2, I used a "combinatorics" approach. Probably slower, but much simpler.


manwar:
Thanks @robbie-hatley

PR #8110, opened 2023-05-21
Added solutions by Simon Dueck in F#

simon-dueck:
Added solutions for task 1 and 2 for week 215, task 1 for week 216 and task 1 for week 217


manwar:
Thanks @simon-dueck

PR #8121, opened 2023-05-21
got stuck on task 2, was tired, couldn't think straight

dcw803:
hacked a non-optimal solution together, will think again and do C translations tomorrow


manwar:
Thanks @dcw803

PR #8222, opened 2023-06-12
Optimize 220/2

choroba:
If we wrap from both sides, we don't have to start from all the elements, a single one is enough.


manwar:
Thanks @choroba

PR #8273, opened 2023-06-23
Add polettix's solution to challenge-222

polettix:
Hi! Here is my contribution for this week's challenge. 222 what a number, uh?!?


manwar:
Thanks @polettix

PR #8393, opened 2023-07-17
Challenge 226 solutions in Perl & Raku by Packy Anderson

packy:


manwar:
Welcome on board and thanks for your first contributions. @packy

PR #8450, opened 2023-07-27
Challenge 227 solutions by Packy Anderson

packy:
With blog posts for both challenge 227 and 226 in the README files


manwar:
Thanks @packy

PR #8471, opened 2023-07-30
feat: add solution for challenge 227 from BarrOff

BarrOff:
Hello

This pull request adds my Raku solution for task one of challenge 227.


manwar:
Thanks @BarrOff

PR #8596, opened 2023-08-21
Simon's solution to challenge 230

simongreen-net:
Better late than never :)


manwar:
Thanks @simongreen-net

PR #8611, opened 2023-08-22
Challenge 231 by Andreas Vögele

voegelas:
BTW, I don't know whether texts such as https://theweeklychallenge.org/blog/perl-weekly-challenge-231/#TASK2 with "phone" and "sex" in the same sentence are ranked high by search engines 😀 "Gender" might be better and will also make SJWs happy.


manwar:
Thanks @voegelas

PR #8628, opened 2023-08-27
Simon's solution to challenge 231

simongreen-net:
Not late this week :)


manwar:
But I  was too late, sorry. Thanks @simongreen-net

PR #8641, opened 2023-09-01
Remove empty dirs

rcmlz:
Perhaps you want to clean up a bit


manwar:
@rcmlz very good idea.
I have been thinking of doing so for long time but never got around to it.
I will take a closer look before merge it.

Thanks for your contributions.

PR #8780, opened 2023-09-28
Challenge-236 - "Exact Change" and "Count Loops"

demerphq:
Exact Change - use a simple algorithm to manage the wallet. The problem description makes it seem like there might be a need to recursively try multiple solutions, but in practice this is not necessary.

Count Loops - I originally implemented this using a destructive approach with a hash, but in my benchmarks this was slow and also non-determinisitc, and making it deterministic would make it slower. Using an array based approach seems like it would be worse, but in practice, with randomly generated input data a pure array approach was faster. *Shrug*

### DIFF
--- a/challenge-236/demerphq/perl/ch-1.pl
+++ b/challenge-236/demerphq/perl/ch-1.pl
@@ -1,0 +1,60 @@
+use strict;
+use warnings;
+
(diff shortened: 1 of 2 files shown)


manwar:
Thanks @demerphq

PR #8784, opened 2023-09-30
Solutions to challenge 236

jo-37:
Vacation challenge from Cascais, Portugal,
@E7-87-83: CY, thanks for Math::Permutation


manwar:
Thanks @jo-37

PR #8787, opened 2023-10-01
Added blog.txt with link to the blog for Task 2

deoac:


manwar:
Kudos for your efforts. I put blog.txt back in the root folder as the parser for the graphs expected it to be there. @deoac

PR #8792, opened 2023-10-01
Add polettix's solution to challenge-229

polettix:
Hi! Long time no see, here's my contribution after some time of relax.

I hope everything is fine with you :smile: Thanks for being the rock that you are!


manwar:
What a pleasant surprise to see you back. I really missed your daily blogs. Having said, I am happy you got the time to relax. @polettix

PR #8822, opened 2023-10-07
Move solution for #236 from directory for #229

polettix:
Hi! I did a mess with the sparse-checkout last week and put the solution for WC236 inside the directory for WC229.

This merge request aims at moving the solution in the right directory.

Thanks and... apologies for all the trouble.


manwar:
Thanks @polettix

PR #8900, opened 2023-10-18
razetime challenge 239 raku and prolog solutions

razetime:
Hi, new contributor here. Hope the format given is ok. Wanna practice a little raku, so figured i would join in.


manwar:
Welcome onboard !!!
Thanks for our first contributions. @razetime

PR #8903, opened 2023-10-19
Improved consistency check in "ch-2.pl".

robbie-hatley:


manwar:
Thanks @robbie-hatley

PR #8922, opened 2023-10-22
feat: add solutions for challenge 239 from BarrOff

BarrOff:
Hello

This pull request adds my Julia, Nim, Perl and Raku solutions for challenge 239.


manwar:
Thanks @BarrOff

PR #8966, opened 2023-10-29
feat: add solutions for challenge 240 from BarrOff

BarrOff:
Hello

This pull request adds my Julia, Nim, Perl and Raku solutions for challenge 240.


manwar:
Thanks @BarrOff

PR #9023, opened 2023-11-07
Add solutions to 242: Missing Numbers & Flip Matrix by E. Choroba

choroba:
In Missing Numbers, I added the [] for no missing elements. The added test case shows why I think it's needed.


manwar:
Thanks @choroba

PR #9049, opened 2023-11-12
feat: add solutions for challenge 242 from BarrOff

BarrOff:
Hello

This pull request adds my Julia, Nim and Raku solutions for challenge 242.


manwar:
Thanks @BarrOff

PR #9077, opened 2023-11-16
Updated solutions by Robbie Hatley for Weekly Challenge 243.

robbie-hatley:
Ranged loops instead of 3-part loops; ch-1.pl now prints both indices _and_ values; and some minor typo fixes.


manwar:
Thanks @robbie-hatley

PR #9201, opened 2023-12-05
Not an one-liner on ch2!

massa:


massa:
And a little bit of math with matrices and linear systems ;-)

PR #9245, opened 2023-12-16
Challenge 247, Adding Perl and Bash (ch-2) solutions. By Nelo Tovar

ntovar:
Perl and Bash (ch-2) solutions. By Nelo Tovar


manwar:
Thanks @ntovar

PR #9309, opened 2023-12-28
Challenge 249. Add Perl and Bash solutions. By Nelo Tovar

ntovar:
Perl and Bash Solutions by Nelo Tovar


manwar:
Thanks @ntovar

PR #249, opened 2019-06-14
Solutions to Challenge 012

bracteatus:
Solutions to Challenge 1 and 2. Opt-out optional Challenge 3.

PR #9312, opened 2023-12-31
Robbie Hatley's solutions in Perl for The Weekly Challenge #249.

robbie-hatley:
Interesting set of tasks. Turns out, Task 2 has multiple compliant permutations for most strings.


manwar:
Thanks @robbie-hatley

PR #9316, opened 2023-12-31
Simon's solution to challenge 249

simongreen-net:
No blog this week. Got struck down with the 'rona :(


manwar:
Sorry to hear about that. Hope and wish you get better soon. Thanks @simongreen-net

PR #9335, opened 2024-01-02
Challenge 249 - Committing andrew_mehta folder containing README, a P…

ajmetz:
Challenge 249 - Committing andrew_mehta folder containing README, and a Perl solution to Challenge 1. File has Ch_1.pm name rather than ch-1.pl as is a Modulino. Also requires a recent Object::Pad version and Perl Version and can be edited to stipulate lower version numbers if needed. Additional folders such as lib and t2 and their contents also included.

### DIFF
--- a/challenge-249/andrew_mehta/README
+++ b/challenge-249/andrew_mehta/README
@@ -1,0 +1,1 @@
+Solution by Andrew Mehta.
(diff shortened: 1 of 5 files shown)


manwar:
Welcome on board and thanks for your first contributions. @ajmetz

PR #9337, opened 2024-01-02
Fixed error in 250-2.

robbie-hatley:
One too many brackets: "[[[:alnum:]]". Oopsie! I changed it to "[a-zA-Z0-9]" which is much more clear.


manwar:
Thanks @robbie-hatley

PR #9341, opened 2024-01-02
New year, back on the challenges

Scimon:
Always fun when there's more lines of tests than lines of code.


manwar:
Welcome back and thanks for your contributions. @Scimon

PR #9378, opened 2024-01-09
Solve 251: Concatenation Value & Lucky Numbers by E. Choroba

choroba:
Including a nice latex proof :-)


manwar:
Thanks @choroba

PR #9488, opened 2024-01-29
Solve PWC254

wlmb:
Regards


wlmb:
Looking at other's, I guess there is an ambiguity in the statement of task 1. Some, including myself, interpreted 'power of three' as n^3 for some n (which is consistent with 0==0^3), but others interpreted it as 3^n.

manwar:
Thanks @wlmb

PR #9580, opened 2024-02-13
modified:   challenge-256/mattneleigh/perl/ch-2.pl

mattneleigh:
Changed variable name to be more correctly descriptive; minor documentation tweak to match


manwar:
Thanks @mattneleigh

PR #9590, opened 2024-02-15
Recent Python scripts

zapwai:
Sometimes I also do challenges from the first 190 weeks, i.e. before I had joined.
Should I upload those very old ones, some time?


manwar:
You are most welcome to do past challenges. @zapwai

PR #9717, opened 2024-03-10
feat: add solutions for challenge 259 from BarrOff

BarrOff:
Hello

This pull request adds my Julia and Raku solutions for task one of challenge 259.


manwar:
Thanks @BarrOff

PR #9725, opened 2024-03-11
Challenge 260 Solutions (Raku)

andemark:


jo-37:
@andemark: Our results for the bitstring in task 2 differ by 237. Very strange! I have no idea, where this comes from. Dropping 6 bits leads to equal ranks in both implementations. See my [perl implementation](https://github.com/manwar/perlweeklychallenge-club/blob/master/challenge-260/jo-37/perl/ch-2.pl) and [blog](https://github.sommrey.de/the-bears-den/2024/03/15/ch-260.html#task-2-dictionary-rank)

jo-37:
@andemark: Looks like a precision problem with Raku types. I am a complete Raku noob, but this change makes our results agree:
```raku
        take ([+] $bag{@less-than}) / ([*] $bag.values>>!).FatRat;
```

andemark:
Thanks for pointing that out.

 
I went ahead and reduced the string to 55 bits just to be safe.

 
I really enjoyed your blog and how in depth it went into the theory behind this problem!

 
Thanks again,

 
-Mark

 
-----Original message-----

PR #9756, opened 2024-03-17
Jörg Sommrey pointed out a precision issue with example 5

andemark:
Jörg Sommrey pointed out a precision issue with example 5. Thanks! 😊


manwar:
Thanks @andemark

PR #9783, opened 2024-03-20
Challenge 261 solutions by Packy Anderson

packy:
* Raku that looks like Perl
  * Perl
  * Python that looks like Perl

1 Blog post


manwar:
Just love your sense of humor, thanks @packy

PR #9793, opened 2024-03-22
Correction to my solution to 261-2.

robbie-hatley:


manwar:
Thanks @robbie-hatley

PR #9913, opened 2024-04-12
Pwc264 - Fix Ch-1

pokgopun:
fix missing out a constraint on ch-1 and add additional test to confirm the fix.


manwar:
Thanks @pokgopun

PR #10024, opened 2024-05-01
updated the task2 challenge 267

princyym:
added the challenge 267 task2 which states that " Line Counts" 
and the problem statement is : You are given a string, $str, and a 26-items array @widths containing the width of each character from a to z.

Write a script to find out the number of lines and the width of the last line needed to display the given string, assuming you can only fit 100 width units on a line.


manwar:
Thanks @princyym

princyym:
Your welcome @maanwar

PR #10039, opened 2024-05-05
PHP solutions for weeks 266 and 267

2colours:
Hi,

long time no see! I had a bit of hiatus but didn't forget how useful these challenges can be to think algorithmically and try out the built-ins of various languages.
These are rather loose, "happy path" implementations in PHP but I enjoyed writing them so I hope they have a place. :)

Cheers.


manwar:
Welcome back and thanks for your contributions. @2colours

PR #10062, opened 2024-05-07
Branch for challenge 268

2colours:
Hi,

thank you for the challenges. These are very simplistic, optimistic "underthinking" solutions based on the examples, with the additional constraint that non-offset scenarios will output "N/A" for the first task.

Also, from what I can tell, this is the first solution of Groovy. I couldn't find a "groovy" folder or a ".groovy" or ".gsh" file. I found this a bit surprising at first but there are only 6 ".ts" files so maybe I shouldn't be so surprised if there is no Groovy.


manwar:
Thanks @2colours

PR #10100, opened 2024-05-15
Challenge 269 solutions by Packy Anderson

packy:
  * Perl
  * Elixir that looks like a hybrid between Perl and Elixir

1 Blog post


manwar:
Thanks @packy

PR #10107, opened 2024-05-16
Got rid of temporary in 269-2.

robbie-hatley:


manwar:
Thanks @robbie-hatley

PR #10132, opened 2024-05-21
Solve PWC270

wlmb:
Regards. 
Found a small typo in the examples: a 3 should be 4 in
    'So the total cost would be (1 x $x) + (3 x $y) => (1 x 2) + (4 x 1) => 6'


manwar:
Thanks @wlmb

PR #10135, opened 2024-05-22
Add more robust solution

wlmb:


wlmb:
My previous solution gave the wrong result with some tricky examples

manwar:
Thanks @wlmb

PR #10168, opened 2024-05-27
New solution to challenge 270 task 2

jo-37:
Fixed an issue after reading @wlmb 's [blog](https://wlmb.github.io/2024/05/20/PWC270/)


manwar:
Thanks @jo-37

PR #10213, opened 2024-06-04
Rename and simplify the function in 270/2

choroba:
The old name was wrongly copied from the previous week.


manwar:
Thanks @choroba

PR #10247, opened 2024-06-11
Robbie Hatley's solutions, in Perl, for The Weekly Challenge #273.

robbie-hatley:
Fun with regular expressions, in both cases. :-)


manwar:
Thanks @robbie-hatley

PR #10252, opened 2024-06-12
Solutions for week #270 in PHP

2colours:
Hi,

Week 270 was very interesting: the first task was good for array-based languages, and the second task was good for math dabblers. To my knowledge, this is the best solution to that problem so far:
- it passes all tests of Mr. Choroba
- it tries to give reasonable values for corner cases (short arrays, negative or zero costs) as well
- it is "formula based" in the sense that after we only need the minimum, the maximum, the length and the sum of the array; everything after that has constant-bound complexity, in the spirit of absolute value (un)equalities. **No loops at all**.

I also think that I have proofs for the formulae I'm using, I can give more insights on the solution if somebody is interested. Hopefully it's easier to explain than come up with it; originally it took me two nights, pretty much...

Anyway, thank you very much for this task, this was probably the most interesting journey.

### DIFF
--- a/challenge-270/2colours/php/ch-1.php
+++ b/challenge-270/2colours/php/ch-1.php
@@ -1,0 +1,21 @@
+<?php
+
+$matrix = json_decode(file_get_contents('php://stdin'));
+$solutions = $matrix;
+
+foreach ($matrix as $row => $content) {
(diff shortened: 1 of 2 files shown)


manwar:
Thanks @2colours

PR #10335, opened 2024-06-29
solution #2 task in Prolog

2colours:
Hi,

thank you for the new tasks. Not sure if I'll have time to do the other one but I was playing around with Prolog again recently so here we go...
Cheers


manwar:
Thanks @2colours

PR #10357, opened 2024-07-01
mgo1977 - PWC 276

mgo1977:
Hi,

I'm sending my Perl solution,

regards,
Mariano


manwar:
Welcome on board and thanks for your first contributions. @mgo1977

PR #10374, opened 2024-07-06
Changed for loops in 276-1 from three-part to ranged.

robbie-hatley:


manwar:
Thanks @robbie-hatley

PR #10443, opened 2024-07-15
Challenge 121 part 2 by Jaldhar H. Vyas.

jaldhar:
This is an old one I hadn't completed the first time.  The blog entry has also been updated though the URL is the same.


manwar:
Thanks @jaldhar

PR #10493, opened 2024-07-25
Prolog solutions for week #279

2colours:
Hi,

thank you for the challenges. These are fairly banal, better-than-nothing solutions in Prolog - it's funny how complicated certain very simple problems can be stated, right? :D
Anyway, hopefully this can also add to the diversity of the solutions.
Cheers.


manwar:
Thanks @2colours

PR #10509, opened 2024-07-29
Solve 280: Twice Appearance & Count Asterisks by E. Choroba

choroba:
"Twice Appearance" has two interpretations, show and test solutions for both.


manwar:
Thanks @choroba

PR #10562, opened 2024-08-07
Challenge 281 update by Packy Anderson

packy:
Task 2 in Elixir; remove extraneous newline from Raku/Perl/Python task 2 output.


manwar:
Thanks @packy

PR #10563, opened 2024-08-07
Robbie Hatley's solutions, in Perl, for The Weekly Challenge #281.

robbie-hatley:
Whereas Challenge 281-1 was one of the simplest I've ever worked on, 281-2 was one of the most complex. ("Minimum path" problems are notoriously hard, and some have no closed-form solution.) 

Also, I finally got to put Perl v5.40 to good use in my 281-2 solution, as it uses the new builtin::inf constant.


manwar:
Thanks @robbie-hatley

PR #10574, opened 2024-08-10
Challenge 281. Add Perl (ch-1) and Bash (ch-1) solutions. By Nelo Tovar

ntovar:
Solutions by Nelo R. Tovar


manwar:
Thanks @ntovar

PR #10609, opened 2024-08-13
Add perl solution to challenge 282

pauloscustodio:


manwar:
Thanks @pauloscustodio

PR #10624, opened 2024-08-15
Add Forth solution to challeng 242 task 2

pauloscustodio:


manwar:
Thanks @pauloscustodio

PR #10644, opened 2024-08-18
feat: Raku solution for task 1 of challenge 282 from BarrOff

BarrOff:
Hello

This pull request adds my Raku solution for task one of challenge 282.


manwar:
Thanks @BarrOff

PR #10739, opened 2024-08-31
challenge-283 andrezgz solution

andrezgz:
Hi @manwar, didn't manage to code it in time, but here's my solution to the previous week challenge


manwar:
Better late than never, thanks @andrezgz

PR #10782, opened 2024-09-07
Add Peter Pentchev's Rust solution to 285.

ppentchev:


manwar:
Welcome on board and thanks for your first contributions. @ppentchev

PR #10788, opened 2024-09-08
Add Peter Pentchev's Python solution to 285.

ppentchev:


manwar:
Wow, bonus Python solution. Thanks @ppentchev

PR #10824, opened 2024-09-11
Add Peter Pentchev's Python solutions to 286

ppentchev:
Also fix a couple of things in the other files:
- make the TAP test functions accept an arrayref for the command to run instead of a single string, since we want to run the Python solution using `python3 -B -u /path/to/the/program`
- fix the use of Test::More::skip() in the Self Spammer test function
- fix a shadowed variable in the Order Game test function
- correct the indentation in the documentation's "Implementation details" section
- make the README file a copy of docs/index.md instead of a wrong copy of a completely different project's README file
- point blog.txt to the Ringlet copy of this challenge's documentation


manwar:
Thanks @ppentchev

PR #10832, opened 2024-09-14
Late but better than never (added #283/284/285)

DevSanti12:
Perl solutions for challenges #283 / #284/ #285, thanks!


manwar:
Cool for catching up, thanks @DevSanti12

PR #10883, opened 2024-09-21
✨ Week 78 & 82 in BQN (& 1 APL)

codereport:
@manwar PS - it looks like there is a small typo somewhere (it says Pytho with no N):

image


manwar:
Thanks @codereport

PR #10914, opened 2024-09-27
Simplified block merging in 288-2.

robbie-hatley:


manwar:
Thanks @robbie-hatley

PR #10923, opened 2024-09-29
feat: add solution for challenge 288 from BarrOff

BarrOff:
Hello

This pull request adds my Raku solution for task one of challenge 288.


manwar:
Thanks @BarrOff

PR #10955, opened 2024-10-03
Challenge 289 solutions by Packy Anderson

packy:
* Raku that maybe looks like Raku, but mostly like Perl
  * Perl
  * Elixir that's starting to look like Elixir

1 Blog post


manwar:
Thanks @packy

PR #11021, opened 2024-10-14
:sparkles: Week 291-1 in BQN

codereport:


manwar:
Thanks @codereport

asherbhs:
that `{(𝔽=𝔽⌾⌽)𝕩}` trick is really nice! I didn't find a way to avoid repeating at least one of `⌽` and ``(»+`)`` in my solution, so I'll be using that idea in the future for sure. I think we can even make it neater by writing ``(»+`){𝔽=𝔽⌾⌽}``

manwar:
@asherbhs I wonder what keyboard you use for these symbols. :-)

asherbhs:
@manwar I configure xkb as described [here](https://mlochbaum.github.io/BQN/editors/index.html#xkb-unix), but one day I'd love to have a physical keyboard with the symbols on the keycaps :)

codereport:
> that `{(𝔽=𝔽⌾⌽)𝕩}` trick is really nice! I didn't find a way to avoid repeating at least one of `⌽` and `` (»+`) `` in my solution, so I'll be using that idea in the future for sure. I think we can even make it neater by writing `` (»+`){𝔽=𝔽⌾⌽} ``

Nice, I didn't know you could do this: `` {𝔽=𝔽⌾⌽} `` with out the `𝕩`

PR #11051, opened 2024-10-20
Add solution to challenge 291 task 2

pauloscustodio:


manwar:
Thanks @pauloscustodio

PR #11194, opened 2024-11-22
Solve 296: String Compression & Matchstick Square by E. Choroba

choroba:
Including the bonus: string DEcompression for 296/1.


manwar:
Thanks @choroba

PR #11211, opened 2024-11-24
feat: add solution for challenge 296 from BarrOff

BarrOff:
Hello

This pull request adds my Nim solution for task one of challenge 296.


manwar:
Thanks @BarrOff

PR #11249, opened 2024-12-01
feat: add solutions for challenge 297 from BarrOff

BarrOff:
Hello

This pull request adds my Julia, Nim and Pascal solution for task one of challenge 297.


manwar:
Thanks @BarrOff

PR #11282, opened 2024-12-08
feat: add solutions for challenge 298 from BarrOff

BarrOff:
Hello

This pull request adds my Julia, Nim and Pascal solutions for task two of challenge 298.


manwar:
Thanks @BarrOff

PR #11405, opened 2025-01-05
feat: add solutions for challenge 302 from BarrOff

BarrOff:
Hello

This pull request adds my Julia, Nim, Pascal and Raku solutions for challenge 302.


manwar:
Thanks @BarrOff

PR #11415, opened 2025-01-07
Fixed 1st solution of challenge-302 after getting hint from 'jo-37'

mahnkong:
Fixed ch-1.pl based on a hint from @jo-37 (https://github.com/manwar/perlweeklychallenge-club/pull/11384#issuecomment-2573281919)


manwar:
Thanks @mahnkong

PR #11420, opened 2025-01-09
New solution to challenge 299 task 2

jo-37:
Improved solution that can handle larger grids.
Thanks @choroba for your solution and the permission to use it as benchmark and @moocow-the-bovine for your support on `PDL::CCS`. Here is the source for moocow-the-bovine/PDL-CCS#14!
See [blog](https://github.sommrey.de/the-bears-den/2025/01/09/ch-299.html) for details.


manwar:
Thanks @jo-37

PR #11437, opened 2025-01-13
Prolog solutions for week #303

2colours:
Hello,

thank you for the challenges, quite funny to play around with Prolog to solve them. One can always learn something new from them, that much is certain haha.

I'm a bit late but please take these solutions for last week's challenges.

Cheers.


2colours:
Right - the solution of the second part is under the assumption that one _must_ clear the whole list. It wouldn't be more difficult to do it otherwise - one would simply drop the potential =< 0 values.

PR #11610, opened 2025-02-20
Week #309 Prolog solutions added

2colours:
Hi,

thank you for the tasks, this really is just a simplistic Prolog solution because I felt like doing it. It's easy to push something interesting in some "alien" language always. :)

Cheers


manwar:
Thanks @2colours

PR #11616, opened 2025-02-23
initial commit of Perl solutions

adamcrussell:
glad to have a chance to submit solutions again after a long break!


manwar:
Thanks @adamcrussell

PR #11624, opened 2025-02-24
LK Perl and Python solutions for Challenge 310

LubosKolouch:
- Implemented arrays intersection function in Perl and Python
- Created sort odd/even indices function in Perl and Python
- Added comprehensive unit tests for both solutions


manwar:
Thanks @LubosKolouch

PR #11719, opened 2025-03-16
sgreen solutions to challenge 311 and 312

simongreen-net:
Double the trouble :)


manwar:
Thanks @simongreen-net

PR #11744, opened 2025-03-19
Arne Sommer

arnesom:
I have one correction to https://theweeklychallenge.org/blog/perl-weekly-challenge-313/

My blog, in the section "Blogs with Creative Title", is indeed very creatively titled. It is not the correct article.


manwar:
Ooops sorry, my mistake.
I will update it now, thanks @arnesom

PR #12134, opened 2025-06-06
challenge 324 idiomatic perl solutions by ysth

ysth:
also added the challenge 323 solutions that I did but forgot to send


manwar:
Thanks @ysth

PR #12218, opened 2025-06-22
feat: add solution for challenge 325 from BarrOff

BarrOff:
Hello

This pull request adds my Raku solutions for task one of challenges 325 and 326.


manwar:
Thanks @BarrOff

PR #12273, opened 2025-07-02
feat(328): completed both challenges in rust

fer2o3:


manwar:
Welcome aboard and thanks for your first contributions. @fer2o3

PR #12344, opened 2025-07-14
Challenge accepted

Scimon:
I added in some tests for the case where the digit is at the start of the string. 
From my reading of the task it should still be deleted. 

(And then while writing this comment I realised I'd overcomplicated it)


manwar:
Thanks @Scimon

PR #12366, opened 2025-07-18
week 330 Arne Sommer

arnesom:
The first week with a new 100% automated git script, and things did not work 100% the first time....


manwar:
Thanks @arnesom

PR #12406, opened 2025-07-24
Blog entries for challenges 001-008 by Jaldhar H. Vyas.

jaldhar:
I discovered that I had written blog entries for the first 8 challenges but apparently never uploaded them.


manwar:
Thanks @jaldhar

PR #12464, opened 2025-08-04
Challenge 333

Scimon:
I'm probably going to add a second MAIN function to the first one I just didn't get around to it.

But Vector classes are fun right?


manwar:
Thanks @Scimon

PR #12487, opened 2025-08-09
Perl solution for task 1 of challenge 333

jkstill:
Solve the collinear problem for both the challenge data, and extended data.


manwar:
Welcome aboard and thanks for the first contributions. @jkstill

PR #12499, opened 2025-08-11
Challenge 334

Scimon:
I may add some more to it but it solved the problems.

I may also get round to blogging this and last week.


manwar:
Thanks @Scimon

PR #12537, opened 2025-08-18
Challange accepted.

Scimon:
I'm kinda thinking about expanding part 2. I completed the challenge but it's so close to being a full noughts and crosses game :)


manwar:
Thanks @Scimon

PR #12631, opened 2025-09-06
Fix solution to task 2

jo-37:
Spied on other solutions to find out that mine was incomplete. This is called _cheating_.


manwar:
Thanks @jo-37

PR #12652, opened 2025-09-09
Challenge 338 solutions by Packy Anderson

packy:
* Raku that is finally Raku-ish
  * Perl
  * Python that kinda looks like Raku

1 blog post


manwar:
Thanks @packy

PR #12659, opened 2025-09-09
Add solutions for Perl Weekly Challenge 334.

LubosKolouch:
Task 1 (Range Sum): Computes the sum of integers between given indices
(inclusive) using array slicing in Perl and Python, with input
validation.

Task 2 (Nearest Valid Point): Finds the index of the valid point
(sharing x or y coordinate) with the smallest Manhattan distance,
returning -1 if none exist, using iteration and distance comparison.


manwar:
Thanks @LubosKolouch

PR #12754, opened 2025-09-29
Solve Perl Weekly Challenge: Broken Keyboard and Reverse Prefix

LubosKolouch:
Implement solutions for Task 1 (Broken Keyboard) and Task 2 (Reverse
Prefix) in both Perl and Python.


manwar:
Thanks @LubosKolouch

PR #13986, opened 2026-04-15
Challenge 369 solutions by Packy Anderson

packy:
* Raku that is kinda Raku-ish
  * Perl
  * Python that kinda looks like Perl

1 blog post with a musical reference


manwar:
Thanks @packy

PR #13988, opened 2026-04-15
Add Perl and C solutions to challenge 369

pauloscustodio:


manwar:
Thanks @pauloscustodio

PR #13996, opened 2026-04-17
Add missing blog.txt for Challenges 366 and 331

LubosKolouch:
Added missing blog.txt for Challenges 366 and 331.


manwar:
Thanks @LubosKolouch

PR #14026, opened 2026-04-22
Add C and Perl solutions to challenge 370

pauloscustodio:
Add C solution to challenge 163


manwar:
Thanks @pauloscustodio

PR #14048, opened 2026-04-27
Add Perl and Python solutions for Challenge 371

LubosKolouch:
Added Perl and Python solutions for Challenge 371.


manwar:
Thanks @LubosKolouch

PR #14094, opened 2026-05-05
Branch for challenge 372

2colours:
Hi,

thank you for the challenges, seems like I'm catching up. These kind of tasks are a nice balance for practicing Prolog - not exactly trivial but not particularly hostile to a declarative, low-state approach. I hope there will be more tasks like that.

Please merge, and have a nice day.


manwar:
Thanks @2colours

PR #14100, opened 2026-05-06
Challenge 372

jchamml:
Added Perl and one-liner solutions for challenge 372.  Updated 370.  One solution for 371.


manwar:
Thanks @jchamml

PR #14103, opened 2026-05-07
Branch for challenge 312

2colours:
Hi,

I found some old task again, it didn't seem to difficult, and then it was only logical to submit it. Please merge. Have a nice afternoon!


manwar:
Thanks @2colours

PR #14151, opened 2026-05-17
Branch for challenge 373

2colours:
Hi,

thank you for the challenges. Overkilled the second task a little bit but that's somewhere along the lines the point of doing these challenges so I'd say Mark nailed it just fine. :)

Have a nice weekend yet.


manwar:
Thanks @2colours

PR #14157, opened 2026-05-18
Add solutions by Andrea Piseri.

ap29600:
working with the k4 interpreter at https://codeberg.org/growler/k


manwar:
Thanks @ap29600

PR #14229, opened 2026-06-01
Branch for challenge 71

2colours:
Hi,

I think this older solution will have to do instead of this week. Actually, I made it last week but it comes in handy. :D (I hate to say it but that "regex task" is is underdefined in significant ways - one cannot just say "find content in HTML tags as well" when HTML tags compose in ways one cannot even match with regular expressions and writing a proper HTML parser is totally out of the league of the usual tasks.)
Please merge. Cheers.

### DIFF
--- a/challenge-071/2colours/README
+++ b/challenge-071/2colours/README
@@ -1,0 +1,1 @@
+Solution By Márton Polgár.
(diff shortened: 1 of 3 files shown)


manwar:
Thanks @2colours

PR #14266, opened 2026-06-08
Solve PWC377

wlmb:
Hi,
I guess the third example of task 2 has some ambiguity: if s1 is equal to s2, the s1 is prefix and suffix of s2 and viceversa and I get the count 2, not 1. If we accept 1 as the correct answer, I believe it is unclear what should be the answer for three or more equal strings?
Best regards,
Luis


manwar:
Thanks @wlmb

PR #14271, opened 2026-06-10
Challenge 376 by Jaldhar H. Vyas.

jaldhar:
Sorry a bit late this week.


manwar:
Better late than never, thanks as always. @jaldhar

PR #14504, opened 2026-07-26
Prolog solution and blog

adamcrussell:


manwar:
This is what I call complete solution, thanks @adamcrussell

PR #14511, opened 2026-07-26
Simplified 383-2.

robbie-hatley:
Too much repetition in my 383-2, so I made a loop instead.


manwar:
Thanks @robbie-hatley

PR #14660, opened 2026-08-28
Challenge 388 solutions by Packy Anderson

packy:
* Raku that is kinda Raku-ish
  * Perl
  * Python that kinda looks like Perl

1 blog post with a musical reference

🎶 Now I know we had no money, but I was rich as I could be 🎶
🎶 With my code of many colors that my mama taught to me 🎶


manwar:
Thanks @packy

PR #14748, opened 2026-09-16
Add modern Perl solutions for Challenge 036

LubosKolouch:
- Add modern Perl solution for Task 1 (VIN Validation) with checksum calculation and Test::More
- Add modern Perl solution for Task 2 (Knapsack Problem) with combinatorial solver and Test::More


manwar:
Thanks @LubosKolouch

